Issue #96, opened 2021-04-01
Automating docker push when a new version is published


AlessandroMinoccheri:
It could be useful to automate the `docker push` when a new version of `phparkitect` is published.
With our GitHub actions it could be possible I think.

At the moment you need manually to build new image with new tag like this

```
docker image build -t phparkitect/phparkitect:0.1.0 .
```

And push it.

```
docker push phparkitect/phparkitect:0.1.0
```

Same thing for latest version:

```
docker image build -t phparkitect/phparkitect:latest .
docker push phparkitect/phparkitect:latest
```

AlessandroMinoccheri:
closed by #99

PR #99, opened 2021-04-02
Added docker automation on github actions workflow

AlessandroMinoccheri:
Related to: #96 

In this PR I have added the automation to publish docker image.

Expected behavior: 
when a push is done on master, it will automatically push docker image with tag latest
when a new tag is created, it will automatically push docker image with the same tag.


.github/workflows/publish_docker_images.yml
```diff
@@ -0,0 +1,45 @@
+name: ci
+
+on:
+  push:
+    branches:
+      - 'master'
+    tags:
+      - '*'
+  pull_request:
```
micheleorselli, line 9: maybe we could omit that, since we do want to push a new image on two cases:
1) a push on master
2) a tag

what do you think?

.github/workflows/publish_docker_images.yml
```diff
@@ -0,0 +1,45 @@
+name: ci
```
micheleorselli, line 1: should we use more descriptive names for out workflows?

.github/workflows/publish_docker_images.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: ${{ github.event_name != 'pull_request' }}
```
micheleorselli, line 43: Should the push happen only it the build is green i.e when all the checks in the workflow "build" pass?

.github/workflows/publish_docker_images.yml
AlessandroMinoccheri: Yes, it could be good enough at the moment 👍

.github/workflows/publish_docker_images.yml
AlessandroMinoccheri: What do you suggest? Same as the filename `publish_docker_images`?

micheleorselli: how does "publish image on docker hub" sounds?

.github/workflows/build.yml
```diff
@@ -40,3 +44,35 @@ jobs:
 
     - name: Test
       run: ./bin/phpunit
+
+  publish_docker_images:
+    needs: build
+    runs-on: ubuntu-latest
+
+    if: github.ref == 'refs/heads/master' || github.event_name == 'release'
```
AlessandroMinoccheri, line 52: @micheleorselli we need to filter actions only for: pushes on master and new tags.

.github/workflows/build.yml
```diff
@@ -40,3 +44,35 @@ jobs:
 
     - name: Test
       run: ./bin/phpunit
+
+  publish_docker_images:
+    needs: build
```
AlessandroMinoccheri, line 49: @micheleorselli If the job named `build` passes, this task can starts.

micheleorselli (approved):
Looks good to me! 👌

PR #103, opened 2021-04-10
downgrading supported php version to php 7.1

AlessandroMinoccheri:
In this PR there is the downgrade to PHP7.1 to support more projects.


micheleorselli (approved):
Looks good to me 🚀!

Issue #98, opened 2021-04-01
Add verbose mode


AlessandroMinoccheri:
It could be useful to add a verbose mode because sometimes is necessary to see which file phparkitect is parsing and if there are some problems.

micheleorselli:
I think we could provide a couple of ways to display progress
- a compact mode, where a progress bar from 0 to 100% is showed
- a verbose mode, where every analysed file is listed

Will this be enough as a first step?

AlessandroMinoccheri:
Yes, each way can be a different pull request, what do you think?

PR #104, opened 2021-04-11
Progress output see #98

micheleorselli:
The PR is still a WIP (fixes https://github.com/phparkitect/arkitect/issues/98) , though enough to get feedback, comments below
There is a major change about the runner, switching from a callback based approach to a plain old iterative way
The code still needs to be cleaned up


src/Analyzer/FileVisitor.php
```diff
@@ -45,12 +47,24 @@ public function onClassAnalyzed(callable $callable): void
...
+        $this->classDescriptions = [];
+    }
+
     public function leaveNode(Node $node): void
     {
         if ($node instanceof Node\Stmt\Class_) {
             $classDescription = $this->classDescriptionBuilder->get();
 
-            \call_user_func($this->callable, $classDescription);
+            $this->classDescriptions[] = $classDescription;
```
micheleorselli, line 65: instead of relying on a callback we accumulate class description here

src/CLI/Check.php
```diff
@@ -41,7 +41,12 @@ protected function configure(): void
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         ini_set('memory_limit', '-1');
+
         try {
+            $verbose = $input->getOption('verbose');
+
+            $progress = $verbose ? new DebugProgress($output) : new ProgressBarProgress($output);
```
micheleorselli, line 48: DebugProgress shows every file parsed
ProgressBarOutput shows a progress bar

src/CLI/Progress.php
```diff
@@ -0,0 +1,17 @@
+<?php
+declare(strict_types=1);
+
+namespace Arkitect\CLI;
+
+use Arkitect\ClassSet;
+
+interface Progress
```
micheleorselli, line 8: every steps method in this interface represents a different steps in analysis. Concrete implementations can decide which information to show and how

src/Analyzer/FileVisitor.php
AlessandroMinoccheri: Make sense this approach

AlessandroMinoccheri:
For me, the approach and the code are correct.
I think it could be useful to add a little section or a description of the new option `-v` inside the README file.

micheleorselli:
> For me, the approach and the code are correct.
> I think it could be useful to add a little section or a description of the new option `-v` inside the README file.

yup, I'll proceed to complete the cleanup and add some info on the README

src/CLI/ProgressBarProgress.php
```diff
@@ -0,0 +1,45 @@
...
+    public function endParsingFile(string $file): void
+    {
+        $this->progress->advance();
+    }
+
+    public function endFileSetAnalysis(ClassSet $set): void
+    {
+        $this->progress->finish();
+        $this->output->writeln('');
+        $this->output->writeln('');
```
AlessandroMinoccheri, line 43: Are necessary these empty writelines?

micheleorselli: Without them, any output would be in the same line of the progress bar

AlessandroMinoccheri: Ok, now it's clear!

AlessandroMinoccheri (approved):
LGTM 👍

Issue #108, opened 2021-04-14
[BUG] Parse Error: Syntax error, unexpected T_MATCH


AlessandroMinoccheri:
When you invoke a class named `Match` PhpParser throw an error:

```
Parse Error: Syntax error, unexpected T_MATCH on line 10#0 /arkitect/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php(158): PhpParser\ParserAbstract->doParse()
#1 /arkitect/vendor/nikic/php-parser/lib/PhpParser/Parser/Multiple.php(51): PhpParser\ParserAbstract->parse('<?php\ndeclare(s...', Object(PhpParser\ErrorHandler\Throwing))
#2 /arkitect/vendor/nikic/php-parser/lib/PhpParser/Parser/Multiple.php(32): PhpParser\Parser\Multiple->tryParse(Object(PhpParser\Parser\Php7), Object(PhpParser\ErrorHandler\Throwing), '<?php\ndeclare(s...')
...
```

The problem is the name `Match` because in PHP8 is a reserved key.
We need to specify that we are using PHP7.1 for example.

AlessandroMinoccheri:
Closed by #109

PR #109, opened 2021-04-14
fixed bug T_MATCH on parsing file

AlessandroMinoccheri:
Refer to #108 

I have explicit that we are using php7.1 when PhpParser is instantiated.


tests/E2E/CliTest.php
```diff
@@ -98,6 +98,22 @@ public function test_bug_yield(): void
         $this->assertStringContainsString($expectedErrors, $process->getOutput());
     }
 
+    /**
+     * @group 8.x
```
micheleorselli, line 102: phpunit has a way to exclude tests based on phpunit version: https://phpunit.readthedocs.io/en/7.3/incomplete-and-skipped-tests.html#skipping-tests-using-requires

This would result in a simpler build configuration, since we could avoid checking for php version here: `run: ./bin/phpunit ${{ matrix.php-versions == '8.0' && ' --exclude-group 8.x' || '' }}`

.php_cs.dist
```diff
@@ -1,6 +1,7 @@
 <?php
 
 $finder = PhpCsFixer\Finder::create()
+  ->exclude([__DIR__.'/tests/E2E/fixtures/mvcMatchBug'])
```
micheleorselli, line 4: though not mandatory, we do have a way to create test fixtures using a virtual file system, eg: https://github.com/phparkitect/arkitect/blob/master/tests/Unit/ClassSetTest.php#L69

This could avoid creating a lot of files or directories, just for testing purpose

AlessandroMinoccheri: Yes, It's better to create test fixtures using a virtual file system. I agree with you. 
Thanks.

tests/E2E/CliTest.php
AlessandroMinoccheri: Right! I can try to use this approach.

.php_cs.dist
AlessandroMinoccheri: We can't use a virtual file system because with integration tests it can't work at the moment.

src/Analyzer/FileParser.php
```diff
@@ -3,12 +3,14 @@
...
 namespace Arkitect\Analyzer;
 
+use PhpParser\Lexer\Emulative;
 use PhpParser\NodeTraverser;
 use PhpParser\NodeVisitor\NameResolver;
 use PhpParser\ParserFactory;
 
 class FileParser implements Parser
 {
+    private const PHP_VERSION = '7.1';
```
AlessandroMinoccheri, line 13: @pfazzi @nelsonkopliku @micheleorselli @fain182 
Here I have hardcoded this PHP version for the parser.
I think that for now is good enough but I would like to open a new issue/pull request if it's a problem.
What do you think?

.github/workflows/build.yml
```diff
@@ -12,10 +12,18 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-versions: [ '7.1', '8.0' ]
```
fain182, line 17: we have a fast build, what about adding even 7.2, 7.3 and 7.4?

src/Analyzer/FileParser.php
fain182: Good enough for me, maybe i would make the name more explicit, it's a PHP version, but why is used the 7.1?

fain182 (approved):

.github/workflows/build.yml
AlessandroMinoccheri: I agree with you!

src/Analyzer/FileParser.php
AlessandroMinoccheri: Yes, I think that we need to be sure for this version and why .

.github/workflows/build.yml
```diff
@@ -12,10 +12,18 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-versions: [ '7.1', '7.2', '7.3', '7.4', '8.0' ]
```
AlessandroMinoccheri, line 17: @fain182 Added more PHP versions with success and it's very fast

micheleorselli:
Looks good to me! 🔝

Issue #107, opened 2021-04-14
Downgrade dependencies


AlessandroMinoccheri:
I think it could be useful to downgrade our dependencies to the version that can support PHP 7.1  and support Symfony projects that are using a version minor than Symfony 4.4  (example Symfony 3.4)

pfazzi:
My 2 cents: this example IMHO confirms how important it is to find an alternative way to install the tool.

PS: If we downgrade we still have to support later versions as well, and if for instance Symfony Console version 6 will not be backwards compatible, the game will no longer work.

AlessandroMinoccheri:
I agree with you @pfazzi we need to find a good strategy for install the tool.

AlessandroMinoccheri:
Closed by #110

PR #110, opened 2021-04-19
Support multiple Symfony versions

AlessandroMinoccheri:
Related to #107

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,16 +23,16 @@
     ],
     "require": {
         "php": "^7.1|^8",
-        "symfony/finder": "^4.4",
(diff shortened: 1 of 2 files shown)


micheleorselli:
The code looks good to me, do you think we need to try if everything runs fine with the older packages versions?

AlessandroMinoccheri:
@micheleorselli I think it can be useful, we should also try to edit our CI introducing older packages versions?

AlessandroMinoccheri:
It seems that with a Symfony 3.1 for example, `composer install` works fine 👍

PR #112, opened 2021-04-24
Added automation to create  phar file

AlessandroMinoccheri:
It creates phar file when a new release is created.


.github/workflows/build.yml
```diff
@@ -84,3 +84,44 @@ jobs:
...
+    needs: build
+    runs-on: ubuntu-latest
+
+    if: github.event_name == 'release'
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: 7.4
```
micheleorselli, line 98: I suggest to change this to php 7.1, since 7.1 is the lowest version we decided to support. How does that sound?

.github/workflows/build.yml
```diff
@@ -84,3 +84,44 @@ jobs:
...
+        restore-keys: |
+          ${{ runner.os }}-php-
+
+    - name: Install dependencies
+      #if: steps.composer-cache.outputs.cache-hit != 'true'
+      uses: php-actions/composer@v5
+      with:
+        args: --prefer-dist
+        composer_version: 2
+        php_version: 7.4
```
micheleorselli, line 118: same as above

.github/workflows/build.yml
AlessandroMinoccheri: I got problems with the build with 7.1 but I could retry to understand well the problem.

AlessandroMinoccheri: @micheleorselli this is the reason:
https://github.com/phparkitect/arkitect/runs/2498698001?check_suite_focus=true

Have you ever seen that error?

AlessandroMinoccheri: It seems fixed with this additional parameter into the configuration:

`"dump-autoload": false`

micheleorselli:
the code looks good to me!

PR #123, opened 2021-05-14
removed version from output cli command

AlessandroMinoccheri:


src/CLI/Check.php
```diff
@@ -96,7 +96,7 @@ protected function printSummaryLine(OutputInterface $output, int $assertionCount
 
     protected function printHeadingLine(OutputInterface $output): void
     {
-        $output->writeln("<info>PHPArkitect 0.0.1</info>\n");
+        $output->writeln("<info>PHPArkitect</info>\n");
```
micheleorselli, line 99: What is the idea behind this change? Imo we should keep it and adding some more infos (a la composer).

AlessandroMinoccheri: Yes, we can do this.
My thought was to do a little step because now it's wrong that information.
We can after adding to composer.json the version information.

Or do you prefer to do this in a single PR?

AlessandroMinoccheri: @micheleorselli I am watching other tools and it seems that no one takes the version from other parts.

Example from PHPUnit

https://github.com/sebastianbergmann/phpunit/blob/master/src/Runner/Version.php

They have hardcoded the version and when. they prepare a new release they change that string.

We can keep our code but we need to change it with the next release.
Previously I thought that was a bad practice but now I don't think so.

What do you think?

fain182: https://stackoverflow.com/questions/50621578/get-version-from-composer-json-symfony/50644633
I think having the version just in one place is a good idea, and composer.json seems the most obvious one..

micheleorselli: +1 for using composer

AlessandroMinoccheri: I have seen that link but it's useful for installed packages not for a custom parameter I think.
Or am I wrong?

@micheleorselli @fain182

AlessandroMinoccheri: I have taken the version from our composer.json file what do you think @micheleorselli @fain182?
Is it good enough?

fain182: Simple and effective 👏

src/CLI/Version.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Arkitect\CLI;
+
+class Version
+{
+    private const COMPOSER_PATH = __DIR__.'/../../composer.json';
```
micheleorselli, line 9: does it also works on the phar version?

tests/Unit/CLI/VersionTest.php
```diff
@@ -0,0 +1,16 @@
...
+namespace Arkitect\Tests\Unit\CLI;
+
+use Arkitect\CLI\Version;
+use PHPUnit\Framework\TestCase;
+
+class VersionTest extends TestCase
+{
+    public function test_it_should_return_version(): void
+    {
+        $this->assertEquals('0.1.11', Version::get());
```
micheleorselli, line 14: This will fail when we bump the version up, is that intended?

src/CLI/Version.php
AlessandroMinoccheri: I have added more possible paths for composer.json file

tests/Unit/CLI/VersionTest.php
AlessandroMinoccheri: Every time we need to update this test. It's not the best solution, we can remove this test if it costs too much in the future.

micheleorselli (approved):

PR #124, opened 2021-05-14
increase psalm level to 3

AlessandroMinoccheri:


src/Analyzer/FileVisitor.php
```diff
@@ -28,12 +28,15 @@ public function enterNode(Node $node): void
...
-        if ($node instanceof Node\Expr\Instanceof_) {
+        if ($node instanceof Node\Expr\Instanceof_ && method_exists($node->class, 'toString')) {
             $this->classDescriptionBuilder
                 ->addDependency(new ClassDependency($node->class->toString(), $node->getLine()));
         }
 
-        if ($node instanceof Node\Expr\New_ && !($node->class instanceof Node\Expr\Variable)) {
+        if ($node instanceof Node\Expr\New_ &&
+            !($node->class instanceof Node\Expr\Variable) &&
+            method_exists($node->class, 'toString')
```
micheleorselli, line 38: Should we extract a method for this particular condition? It find it a bit hard to get

AlessandroMinoccheri: We can't move all conditions in another function for psalm reasons.
But I have improved a little bit.

src/Analyzer/FileVisitor.php
```diff
@@ -63,4 +63,9 @@ public function leaveNode(Node $node): void
             $this->classDescriptions[] = $classDescription;
         }
     }
+
+    private function isNewNodeType(Node $node): bool
```
micheleorselli, line 67: not sure i'm getting where this method is used. Apart from that, I think we should keep track why we are adding such checks and condition

AlessandroMinoccheri: It was a refusal.

micheleorselli (approved):

PR #133, opened 2021-05-21
Adds extend rule

### LINKED ISSUE #128: Adds rule to check class extension
we want to be able to check that all classes extending a base class must follow some rules

micheleorselli:
This PR fixes #128

### DIFF
--- a/src/Analyzer/ClassDescriptionBuilder.php
+++ b/src/Analyzer/ClassDescriptionBuilder.php
@@ -14,6 +14,9 @@ class ClassDescriptionBuilder
     /** @var array */
     private $interfaces;
 
+    /** @var FullyQualifiedClassName */
+    private $extend;
+
     /** @var string */
     private $filePath;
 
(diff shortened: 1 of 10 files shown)


src/Analyzer/ClassDescriptionBuilder.php
```diff
@@ -54,9 +57,17 @@ public function addDependency(ClassDependency $cd): self
         return $this;
     }
 
+    public function setExtends(string $FQCN, int $line): self
```
AlessandroMinoccheri, line 60: At some point is `Extends` in other `Extend`, what do you think to unify the name?

AlessandroMinoccheri (changes requested):

src/Analyzer/ClassDescriptionBuilder.php
micheleorselli: Thing is, extends is a reserved word and cannot be used as class name. I tried to keep consistency inside every class

PR #159, opened 2021-09-14
added naming conventions examples

AlessandroMinoccheri:


examples/naming-conventions.md
```diff
@@ -0,0 +1,25 @@
+# Naming Conventions
+
+In this example, we would like to share some simple rules to apply some naming conventions in your project.
+
+Naming conventions can help the team to create and rename new files with a common shared standard.
```
micheleorselli, line 5: classes, not files

examples/naming-conventions.md
```diff
@@ -0,0 +1,25 @@
+# Naming Conventions
+
+In this example, we would like to share some simple rules to apply some naming conventions in your project.
+
+Naming conventions can help the team to create and rename new files with a common shared standard.   
+
+In the following rules, written into `phparkitect.php` file into the root of a project, we said that:
```
micheleorselli, line 7: That would be the case for every example, should we repeat it every time? 🤔

examples/naming-conventions.md
```diff
@@ -0,0 +1,25 @@
...
+```
+* All files inside namespace App\Infrastructure\Controller should end with Controller
+* All files inside namespace App\Domain\Auction\Bids should start with Bid
```
micheleorselli, line 10: classes, not files

micheleorselli (changes requested):

examples/naming-conventions.md
AlessandroMinoccheri: Right

examples/naming-conventions.md
AlessandroMinoccheri: Right.

examples/naming-conventions.md
AlessandroMinoccheri: ok, do you think to remove all the seven line? 
Do you have some suggestions @micheleorselli ?

Jean85: What about:
```
In this example, the desired rules are:
# ...
Which can be translated in the configuration like this:
```

AlessandroMinoccheri: Sounds good to me! 👍 
@micheleorselli ?

micheleorselli: yup, good to me!

Issue #160, opened 2021-09-18
Crashes with PHP 7.4 syntax


davidstoker:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Library Version | 0.1.14
| PHP version     | 7.4.23

#### Summary

Current PHP-Parser configuration in this project causes crashes within `php-parser` using newer syntax introduced in 7.4.

#### Current behavior

I was trying this out on a large project that runs PHP 7.4. When running an initial check, many crashes dumped out from `php-parser` with `Parse Error: Syntax error, unexpected T_STRING, expecting T_PAAMAYIM_NEKUDOTAYIM on line X`.

The places referenced utilize Arrow functions which were introduced in PHP 7.4.

`FileParser` in this project configures an `Emulative` from `php-parser` with a `phpVersion` of 7.1

https://github.com/phparkitect/arkitect/blob/911bf96da29ae095069680771070f8d36c244e63/src/Analyzer/FileParser.php#L13-L30

The `Emulative` options say `phpVersion` sets the version to emulate which means this is forcing emulation of 7.1? Is this intentional? Commenting out that option that's passed to `Emulative` allows the checks to run correctly on my project.

https://github.com/nikic/PHP-Parser/blob/0a20979a62bc159fa0159ec7a4b5e2360bfa1abc/lib/PhpParser/Lexer/Emulative.php#L38-L40

#### How to reproduce

Try running checks on a file using PHP 7.4+ on a file with an arrow function like `$fn1 = fn($x) => $x + $y;` in it.

#### Expected behavior

Checks would execute without crashing.

AlessandroMinoccheri:
Thanks, @davidstoker for this issue, I am working to fix it in this PR: #161

4n70w4:
May be similar issue:

```
Parse Error: Syntax error, unexpected T_VARIABLE, expecting ':' on line 193#0 phar:///var/www/promo/phparkitect.phar/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php(159): PhpParser\ParserAbstract->doParse()
```

```
Parse Error: Syntax error, unexpected T_DOUBLE_ARROW, expecting ')' on line 325#0 phar:///var/www/promo/phparkitect.phar/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php(159): PhpParser\ParserAbstract->doParse()
```

```
[...]

PR #161, opened 2021-09-18
fix parsing arrow function using php_version dynamically into parser

AlessandroMinoccheri:
In this PR I removed the PHP version fixed for the emulator, and now we are using the PHP version dynamically.
Removing PHP version specification break all the tests.

Related issue: #160

### DIFF
--- a/src/Analyzer/FileParser.php
+++ b/src/Analyzer/FileParser.php
@@ -10,7 +10,6 @@ use PhpParser\ParserFactory;
 
 class FileParser implements Parser
 {
-    private const PHP_VERSION = '7.1';
     /** @var \PhpParser\Parser */
     private $parser;
 
@@ -26,7 +25,7 @@ class FileParser implements Parser
 
         $lexer = new Emulative([
             'usedAttributes' => ['comments', 'startLine', 'endLine', 'startTokenPos', 'endTokenPos'],
-             'phpVersion' => self::PHP_VERSION,
+             'phpVersion' => phpversion('tidy'),
         ]);
 
         $this->parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7, $lexer);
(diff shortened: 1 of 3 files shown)


src/Analyzer/FileParser.php
```diff
@@ -26,7 +25,7 @@ public function __construct(NodeTraverser $traverser, FileVisitor $fileVisitor,
 
         $lexer = new Emulative([
             'usedAttributes' => ['comments', 'startLine', 'endLine', 'startTokenPos', 'endTokenPos'],
-             'phpVersion' => self::PHP_VERSION,
+             'phpVersion' => phpversion('tidy'),
```
micheleorselli, line 28: tidy is not needed

micheleorselli:
looks good to me though I was wondering if this could cause problems 🤔

src/Analyzer/FileParser.php
AlessandroMinoccheri: for the coding standard, we need to use it.

AlessandroMinoccheri:
The problem @micheleorselli could be with the generated `.phar`? At the moment we use 7.1 always, now with this PR the PHP version depends on the PHP installed.
We can add multiple `.phar` files with different PHP versions.
At the moment I can't figure out other problems but we can try arkitect on different projects.

micheleorselli:
Hi,

> The problem @micheleorselli could be with the generated `.phar`? At the moment we use 7.1 always, now with this PR the PHP version depends on the PHP installed.

yeah I think so

> We can add multiple `.phar` files with different PHP versions.

Not really sure we want to do that, we should stick with the minimum php supported version. Still, the phar should be able to work with newver php versions

> At the moment I can't figure out other problems but we can try arkitect on different projects.

src/Analyzer/FileParser.php
AlessandroMinoccheri: You are right, I have removed the extension @micheleorselli

micheleorselli:
We can merge and release this. I guess the same issue potentially applies to php 8 and 8.1. In the meantime we can merge this

PR #165, opened 2021-09-25
added check for method toString

AlessandroMinoccheri:
I have added another check for method `toString` related to this issue: #160

### DIFF
--- a/src/Analyzer/FileVisitor.php
+++ b/src/Analyzer/FileVisitor.php
@@ -51,7 +51,8 @@ class FileVisitor extends NodeVisitorAbstract
         }
 
         if ($node instanceof Node\Expr\New_ && !($node->class instanceof Node\Expr\Variable)) {
-            if (method_exists($node->class, 'isAnonymous') && $node->class->isAnonymous()) {
+            if ((method_exists($node->class, 'isAnonymous') && $node->class->isAnonymous()) ||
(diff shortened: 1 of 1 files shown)


micheleorselli:
What does it fixes? Should we write a test about it?

AlessandroMinoccheri:
This PR should fix:
```
Call to undefined method PhpParser\Node\Expr\PropertyFetch::toString()
```

Do you have an idea how to replicate a `PropertyFetch` Expression to add into our test suite @micheleorselli ?

micheleorselli:
Do we know what snippet is generating the problem?

AlessandroMinoccheri:
Unfortunately no, I asked inside the issue to run Arkitect in verbose mode to understand where is the problem.

For that reason, it's not easy to create a test for it.
But the error seems clear and the if condition added inside this PR could solve the problem.

micheleorselli:
If we are not able to write a test that reproduces the bug we are guessing what the problem is. Since the problem is in the class `PropertyFetch` I tried to look for something similar in the PHP Parser tests. I found something that could be useful: https://github.com/nikic/PHP-Parser/blob/master/test/code/parser/expr/fetchAndCall/objectAccess.test. We could take inspiration from that and write a test which exposes the bug. What do you guys think?

PR #166, opened 2021-09-25
added not expressions and deprecate not logic

AlessandroMinoccheri:


src/Expression/Logical/Not.php
```diff
@@ -9,6 +9,9 @@
 use Arkitect\Expression\NegativeDescription;
 use Arkitect\Rules\Violations;
 
+/**
+ * @deprecated deprecated
+ */
 class Not implements Expression
```
fain182, line 15: Why deprecate and not delete? We found that is misleading, so for me it's better remove it.

fain182:
We need to update the README too, I think..

src/Expression/Logical/Not.php
AlessandroMinoccheri: Yes, this is a BC and I tried a different approach before delete it completely. 
I'll remove it completely.

micheleorselli: As of now Not it is just not working as expected, we should get rid of it

AlessandroMinoccheri: I have Removed the class `Not`

micheleorselli (approved):

PR #170, opened 2021-10-03
removes misleading Expressions class

### LINKED ISSUE #163: Extend() should implement Expression instead of extending from Expresssions
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Library Version | 0.1.15
| PHP version     | 8.0.8

#### Summary

Looking at the `Arkitect\Expression\ForClasses\Implement` and `\Arkitect\Expression\ForClasses\Extend` classes, it would seem that the `Extend` class should have a declaration of

`class Extend implements Expression`

instead of currently

`class Extend extends Expressions`

I have tried this and the analysis still runs fine. (however, I'm still getting a weird result as per my other issue reported in #162 which may or may not be related).

Thanks

fain182:
Fixes #163


micheleorselli (approved):

AlessandroMinoccheri (approved):
LGTM 👍

PR #171, opened 2021-10-06
wip getting extended class dependencies

AlessandroMinoccheri:
In this PR I get all dependencies and interfaces of the extended class.
At the moment it's not recursive but I would like to know if the way is correct or not before adding more effort.


AlessandroMinoccheri:
Closing this PR because we need a different way to parse extended class like an autolader.

PR #178, opened 2021-11-09
Update README.md

ricfio:
I believe that it's better install phparkitect as dev package... or not ?
composer require --dev phparkitect/phparkitect


AlessandroMinoccheri:
Yes @ricfio you are right!
Thanks for your contribution!

Issue #184, opened 2021-11-11
Some tests with assertStringContainsString could fail due Windows EOL (CR+LF) vs Unix EOL (CR)


ricfio:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | yes
| Library Version | 0.2.2 
| PHP version     | 8.0.12

#### Summary

The following PHPUnit tests fail on my WSL2 (Windows 10):
- test_app_returns_error_with_multiple_violations
- test_bug_yield
- test_assertion_should_fail_on_broken_rule

#### Current behavior

Some PHPUnit tests (that use `$this->assertStringContainsString(...)`) fail on WSL2 (Windows 10) due extra LF in `$expectedErrors` (CR+LF) comparing to the actual `$cmdTester->getDisplay()` output (CR only).

The cause of this bug was to clone the repository with my global git config option autocrlf setted to true:
`git config --global core.autocrlf true`

This issue can be solved cloning the repository with core.autocrlf setted to false or converting EOL in the related source files (`tests/E2E/Cli/CheckCommandTest.php` for example) from Windows (CR+LF) to Unix (CR).

I'm not sure, but could be better solve avoid this issue directly with an update of the test source code.
So, if prefer solve without change the test source code you can close this issue, please. 

Here an example of test fail output for test_app_returns_error_with_multiple_violations:

```
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.0.12
Configuration: /home/ricfio/projects/phparkitect/arkitect/phpunit.xml

F                                                                   1 / 1 (100%)

Time: 103 ms, Memory: 12.00 MB

There was 1 failure:

1) Arkitect\Tests\E2E\Cli\CheckCommandTest::test_app_returns_error_with_multiple_violations
Failed asserting that 'PHPArkitect dunno\n
\n
Config file: /home/ricfio/projects/phparkitect/arkitect/tests/E2E/Cli/../_fixtures/configMvc.php\n
\n
analyze class set /home/ricfio/projects/phparkitect/arkitect/tests/E2E/_fixtures/mvc\n
\n
  0/18 [>---------------------------]   0%\n
 18/18 [============================] 100%\n
\n
ERRORS!\n
\n
App\Controller\Foo violates rules\n
  should implement ContainerAwareInterface\n
  should have a name that matches *Controller\n
\n
App\Controller\ProductsController violates rules\n
  should implement ContainerAwareInterface\n
\n
[...]

PR #185, opened 2021-11-11
removed line breaks when we try to compare the output

AlessandroMinoccheri:
I think that these output tests can be replaced with others but at the moment we can solve removing line brakes when we try to compare the output.

Fixing #184


ricfio:
I agree with you, in the future I suggest comparing the output objects of the check instead of the string output

AlessandroMinoccheri:
@ricfio I think that it's better to test the behavior, not the output.
To test the behavior you need other tests.

ricfio:
> @ricfio I think that it's better to test the behavior, not the output. To test the behavior you need other tests.

So do you agree with me ?

PR #188, opened 2021-11-11
fix namespaces match with wildcard for some use cases

ricfio:


ricfio:
Should we improve `FullyQualifiedClassNameTest` (`patternProvider` input) as already done on `ResideInOneOfTheseNamespacesTest` (`shouldMatchNamespacesProvider`) or move the new test cases from `ResideInOneOfTheseNamespacesTest` to `FullyQualifiedClassNameTest` or do nothing else?

Not sure which is the best choice.

ricfio:
I have used this table to convert from ShellPattern to RegExPattern:
https://www.ibm.com/docs/en/aix/7.2?topic=files-pattern-matching-versus-regular-expressions

image

ricfio:
In this PR I have changed the `PatternString::matches` method replacing fnmatch with `preg_match` and using a RegEx pattern 'starts with' `#^...#` instead of a RegEx pattern 'from begin to end' `#^...$#`.

I believe that the first approach 'starts with' `#^...#` is good for namespaces matching, but I'm not sure that is ok for className matching too, in this last one case could be more correctly and restrictive to use the second approach 'from begin to end' `#^...$#` (such as the previous fnmatch function works, I suppose).

At this moment all test pass so If my dubt will be confirmed we should improve the actual test cases to cover such matching differences between namespaces and classNames matching.

src/Analyzer/PatternString.php
```diff
@@ -23,7 +23,7 @@ public function matches(string $pattern): bool
             return true;
         }
 
-        return fnmatch($pattern, $this->value, \FNM_NOESCAPE);
+        return 1 === preg_match('#^'.$this->convertShellToRegExPattern($pattern).'#', $this->value);
```
AlessandroMinoccheri, line 26: This line can be difficult to read, what do you think about moving this condition into a specific function with a specific name to understand well what it does?

ricfio: Sure, done.
Now we could also move the following code inside the new method `startsWithPattern`:
```
if (!$this->containsWildcard($pattern) && str_starts_with($this->value, $pattern)) {
    return true;
}
```

Alternatively, unless for performance reasons, we could also remove the above code because it's already covered by this:
```
1 === preg_match('#^'.$this->convertShellToRegExPattern($pattern).'#', $this->value);
```

What do you think about this ?

ricfio:
I have another question... should be I to increment the version (in this case to 0.2.5) and update the test VersionTest ?

AlessandroMinoccheri (approved):
LGTM 👍

src/Analyzer/PatternString.php
ricfio: So, you prefer leave this code explicit and do not remove it, right?
```
if (!$this->containsWildcard($pattern) && str_starts_with($this->value, $pattern)) {
    return true;
}
```

```
    private function containsWildcard(string $pattern): bool
    {
        return
            str_contains($pattern, '*') ||
            str_contains($pattern, '?') ||
            str_contains($pattern, '.') ||
            str_contains($pattern, '[');
    }
```

Issue #179, opened 2021-11-09
Expression NotHaveDependencyOutsideNamespace(..) detect false positives


ricfio:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Library Version | 0.2.2
| PHP version     | 8.0.11

#### Summary

Expression NotHaveDependencyOutsideNamespace(...) detect false positives in these cases:
- calling a static method using self keyword
- throwing a new Exception

#### Current behavior

```
App\Domain\Assert violates rules
  should not depend on classes outside in namespace App\Domain (on line 14)
  should not depend on classes outside in namespace App\Domain (on line 21)
```

#### How to reproduce

Using this configuration (phparkitect.php):

```
<?php
declare(strict_types=1);

use Arkitect\ClassSet;
use Arkitect\CLI\Config;
use Arkitect\Expression\ForClasses\NotHaveDependencyOutsideNamespace;
use Arkitect\Expression\ForClasses\ResideInOneOfTheseNamespaces;
use Arkitect\Rules\Rule;

return static function (Config $config): void {
    $allClasses = ClassSet::fromDir(__DIR__.'/src');

    $rule_1 = Rule::allClasses()
        ->that(new ResideInOneOfTheseNamespaces('App\Domain'))
        ->should(new NotHaveDependencyOutsideNamespace('App\Domain'))
        ->because('we want protect our domain');

    $config
        ->add($allClasses, ...[$rule_1]);
};
```

to check this class (Assert.php):

```
<?php

declare(strict_types=1);

namespace App\Domain;

use InvalidArgumentException;

final class Assert
{
    public static function arrayOf(string $class, array $items): void
    {
        foreach ($items as $item) {
            self::instanceOf($class, $item);
        }
    }

    public static function instanceOf(string $class, object $item): void
    {
        if (!$item instanceof $class) {
            throw new InvalidArgumentException(
                sprintf('The object <%s> is not an instance of <%s>', $class, get_class($item))
            );
        }
    }
}

```
#### Expected behavior

No Errors

ricfio:
The first false positive (on line 14) not happens replacing 'self' keyword with ClassName Assert:
```
    public static function arrayOf(string $class, array $items): void
    {
        foreach ($items as $item) {
            Assert::instanceOf($class, $item);
        }
    }
```

AlessandroMinoccheri:
Fixed by #197

PR #197, opened 2021-11-16
removed self as a dependency if it s found

AlessandroMinoccheri:
Fixing #179 
cc @ricfio

### DIFF
--- a/src/Analyzer/FileVisitor.php
+++ b/src/Analyzer/FileVisitor.php
@@ -41,17 +41,30 @@ class FileVisitor extends NodeVisitorAbstract
          * @see FileVisitorTest::test_should_returns_all_dependencies
          */
         if ($node instanceof Node\Expr\StaticCall && method_exists($node->class, 'toString')) {
+            if ($this->isSelfOrStaticOrParent($node->class->toString())) {
+                return;
+            }
+
             $this->classDescriptionBuilder
                 ->addDependency(new ClassDependency($node->class->toString(), $node->getLine()));
         }
 
         if ($node instanceof Node\Expr\Instanceof_ && method_exists($node->class, 'toString')) {
+            if ($this->isSelfOrStaticOrParent($node->class->toString())) {
+                return;
+            }
+
             $this->classDescriptionBuilder
                 ->addDependency(new ClassDependency($node->class->toString(), $node->getLine()));
         }
 
         if ($node instanceof Node\Expr\New_ && !($node->class instanceof Node\Expr\Variable)) {
-            if (method_exists($node->class, 'isAnonymous') && $node->class->isAnonymous()) {
+            if (method_exists($node->class, 'isAnonymous')
+                && $node->class->isAnonymous()) {
+                return;
+            }
+
+            if ($this->isSelfOrStaticOrParent($node->class->toString())) {
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/Analyzer/FileVisitor.php
```diff
@@ -41,17 +41,29 @@ public function enterNode(Node $node): void
          * @see FileVisitorTest::test_should_returns_all_dependencies
          */
         if ($node instanceof Node\Expr\StaticCall && method_exists($node->class, 'toString')) {
+            if ('self' === $node->class->toString()) {
```
ricfio, line 44: This code portion is repeated inside the next three if conditions:
```
            if ('self' === $node->class->toString()) {
                return;
            }
            $this->classDescriptionBuilder
                ->addDependency(new ClassDependency($node->class->toString(), $node->getLine()));
```

Could be a good idea refactor it ?

src/Analyzer/FileVisitor.php
```diff
@@ -41,17 +41,29 @@ public function enterNode(Node $node): void
          * @see FileVisitorTest::test_should_returns_all_dependencies
          */
         if ($node instanceof Node\Expr\StaticCall && method_exists($node->class, 'toString')) {
```
ricfio, line 43: Could this condition be expressed implementing a specific method in Node\Expr\StaticCall?

src/Analyzer/FileVisitor.php
```diff
@@ -41,17 +41,29 @@ public function enterNode(Node $node): void
...
+            if ('self' === $node->class->toString()) {
+                return;
+            }
             $this->classDescriptionBuilder
                 ->addDependency(new ClassDependency($node->class->toString(), $node->getLine()));
         }
 
         if ($node instanceof Node\Expr\Instanceof_ && method_exists($node->class, 'toString')) {
```
ricfio, line 51: Could this condition be expressed implementing a specific method in Node\Expr\Instanceof_?

src/Analyzer/FileVisitor.php
AlessandroMinoccheri: I think not because is there is no method toString it can generate an exception

src/Analyzer/FileVisitor.php
AlessandroMinoccheri: I think not because is there is no method toString it can generate an exception

src/Analyzer/FileVisitor.php
ricfio: Hmm, maybe I read better now... so is $node->class a class outside of phparkitect?
Is it a class of the project that phparkitect is checking out?

src/Analyzer/FileVisitor.php
AlessandroMinoccheri: This is the `Node` that the parser has built after parsing the file.
We are using this library to do this:

https://github.com/nikic/PHP-Parser

ricfio: Ok thanks, so you cannot add methods in `Node \ Expr \ StaticCall` and` Node \ Expr \ InstanceOf_` why its are classes of PHP-Parser and so outside of phparkitect, right?

AlessandroMinoccheri: they are part of a syntax tree generated by the external library and you can't add anything.

ricfio:
Could there be the same bug when using the `static` keyword instead of the `self` keyword?

ricfio:
A bigger problem might be that the `static` keyword (as well as `parent`) should be interpreted before to detect its namespace, unlike `self` wich always indicates the same class within which you use `self`.

AlessandroMinoccheri:
I removed `static` also @ricfio .
I think that `parent` could be a different context and not part of this PR :)

ricfio:
Ok but could  `parent` or also `static` refer to a base class outside the namespace in the rule?
I think not, because in this case you already have in your code an import (`use ..`) with an outside namespace for the base class.

AlessandroMinoccheri:
I am trying with some code and we can avoid to add `parent` as your suggestions @ricfio

PR #217, opened 2021-12-11
fix: no param dependency for special type

ricfio:
It solve false positives when the 'self' or 'static' keywords are used as method param type.

### DIFF
--- a/src/Analyzer/FileVisitor.php
+++ b/src/Analyzer/FileVisitor.php
@@ -113,6 +113,10 @@ class FileVisitor extends NodeVisitorAbstract
             return;
         }
 
+        if (method_exists($node->type, 'isSpecialClassName') && ($node->type->isSpecialClassName())) {
+            return;
+        }
+
         if (!method_exists($node->type, 'toString')) {
             return;
         }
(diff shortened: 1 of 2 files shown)


src/Analyzer/FileVisitor.php
```diff
@@ -113,6 +113,10 @@ private function addParamDependency(Node\Param $node): void
             return;
         }
 
+        if (method_exists($node->type, 'isSpecialClassName') && ($node->type->isSpecialClassName())) {
```
ricfio, line 116: When using `self` or `static` as param type the `$node` object is an instance of `PhpParser\Node\Name`, so this condition could be changed (for example) as follow:
`if (($node->type instanceof Node\Name) && ($node->type->isSpecialClassName())) {`

... but I prefer the actual condition that is more generic because the method `isSpecialClassName()` exists in more classes, for example in `PhpParser\Node\Name` and `PhpParser\Node\Identifier`.

What do you think about ?

AlessandroMinoccheri: We can try to use the actual condition and we can understand if we don't have covered well this case ok?

AlessandroMinoccheri (approved):
LGTM 👍 

Thanks!

Issue #209, opened 2021-12-05
implement except(...) clause to improve expressivity


ricfio:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

We may need a new exception clause (not a php exception) to apply to Rule::allClasses() that at this time does not exists (I believe).

As earlier discussed (@micheleorselli)... 
>  Is something we discussed earlier and, at that time, we decided to not include it in the DSL. Main reason: ->except could be interpreted as a way to
> 
> - exclude classes from being evaluated (allClasses -> except)
> - exclude classes from the that clause (that(ResideInOneOfTheseNamespaces) -> except)
> - exclude classes from the should clause (should(NotHaveDependenciesOutside) -> except)

AlessandroMinoccheri:
What do you think about something like this?

```
$rule = Rule::allClasses()
            ->that(new ResideInOneOfTheseNamespaces('App\HappyIsland'))
            ->should(new HaveNameMatching('Happy*'), new RuleException('App\HappyIsland\Foo'))
            ->because("that's what she said");
```

We can add to `should` method a nullable parameter for exceptions (also for `that` method?)

What do you think @micheleorselli @fain182 ?

fain182:
I think thath the selection of the classes should be in the "that" clause...
so I would think something like:
```
$rule = Rule::allClasses()
            ->that(new ResideInOneOfTheseNamespaces('App\HappyIsland'))
            ->andThat(new IsNotOneOf([''App\HappyIsland\Foo'']))
            ->should(new HaveNameMatching('Happy*'))
            ->because("that's what she said");
```

I am not sure about the name `IsNotOneOf`, even if reading it in the that clause sounds good...

AlessandroMinoccheri:
For the selection of the class should be in `that` I agree, but if you want to exclude some dependencies from the rules you need to use it in `should`.

`isNotOneOf` is a better name, thanks.

AlessandroMinoccheri:
@fain182 but what do you think about adding a second parameter to `should`?
Like this:

```
->should(new HaveNameMatching('Happy*'), new isNotOneOf('App\HappyIsland\Foo'))
```

fain182:
It's not too bad to read, but how would you call in the signature the second argument of the should ? 
I am trying to see if this solution can be generalized or not, because if we add a second argument to the "should", it should make sense always...

AlessandroMinoccheri:
[...]

PR #219, opened 2021-12-16
Added exclude function dsl

AlessandroMinoccheri:
In this PR I added the possibility to exclude classes when we parse the codebase.

This is one of the different steps for this issue #209

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,11 +99,11 @@ return static function (Config $config): void {
 By default, a progress bar will show the status of the ongoing analysis.
 
 ## Excluding classes when parser run
-If you want to exclude some classes from the parser you can use the `exclude` function inside your config file like this:
+If you want to exclude some classes from the parser you can use the `except` function inside your config file like this:
 
 ```php
 $rules[] = Rule::allClasses()
-->exclude('App\Controller\BaseController')
+->except('App\Controller\BaseController')
 ->that(new ResideInOneOfTheseNamespaces('App\Controller'))
 ->should(new HaveNameMatching('*Controller'))
 ->because('we want uniform naming');
--- a/README.md
+++ b/README.md
@@ -98,6 +98,17 @@ return static function (Config $config): void {
 
 By default, a progress bar will show the status of the ongoing analysis.
 
+## Excluding classes when parser run
+If you want to exclude some classes from the parser you can use the `exclude` function inside your config file like this:
+
+```php
+$rules[] = Rule::allClasses()
+->exclude('App\Controller\BaseController')
+->that(new ResideInOneOfTheseNamespaces('App\Controller'))
+->should(new HaveNameMatching('*Controller'))
+->because('we want uniform naming');
+```
+
 ## Optional parameters and options
 You can add parameters when you launch the tool. At the moment you can add these parameters and options: 
 * `-v` : with this option you launch Arkitect with the verbose mode to see every parsed file
--- a/README.md
+++ b/README.md
@@ -103,11 +103,13 @@ If you want to exclude some classes from the parser you can use the `except` fun
 
 ```php
 $rules[] = Rule::allClasses()
-->except('App\Controller\BaseController')
+->except('App\Controller\FolderController\*')
 ->that(new ResideInOneOfTheseNamespaces('App\Controller'))
 ->should(new HaveNameMatching('*Controller'))
 ->because('we want uniform naming');
 ```
+
+You can use wildcards or the exact name of a class.
 
 ## Optional parameters and options
 You can add parameters when you launch the tool. At the moment you can add these parameters and options: 
--- a/src/Rules/AllClasses.php
+++ b/src/Rules/AllClasses.php
@@ -11,9 +11,12 @@ class AllClasses implements ThatParser
 {
     /** @var RuleBuilder */
     protected $ruleBuilder;
+    /** @var array */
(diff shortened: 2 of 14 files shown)


src/Rules/ArchRule.php
```diff
@@ -16,15 +17,25 @@ class ArchRule implements DSL\ArchRule
...
         $this->shoulds = $constraints;
         $this->because = $because;
+        $this->classesToBeExcluded = $classesToBeExcluded;
     }
 
     public function check(ClassDescription $classDescription, Violations $violations): void
     {
+        $fqcn = FullyQualifiedClassName::fromString($classDescription->getFQCN());
+
+        if (\in_array($fqcn->toString(), $this->classesToBeExcluded)) {
```
AlessandroMinoccheri, line 35: Here we can use wildcards also not only check for a single class right?
For example you can say that you want to exclude a folder like"`App\Domain\Foo\*`

What do you think @micheleorselli

AlessandroMinoccheri: We can also move this check inside `allSpecsAreMatchedBy`

micheleorselli: Wildcards are fine, we should support them

README.md
```diff
@@ -98,6 +98,17 @@ return static function (Config $config): void {
...
+## Excluding classes when parser run
+If you want to exclude some classes from the parser you can use the `exclude` function inside your config file like this:
+
+```php
+$rules[] = Rule::allClasses()
+->exclude('App\Controller\BaseController')
+->that(new ResideInOneOfTheseNamespaces('App\Controller'))
+->should(new HaveNameMatching('*Controller'))
+->because('we want uniform naming');
+```
```
fain182, line 110: if we add wildcards, we should remember to add them in the documentation

src/Rules/AllClasses.php
```diff
@@ -23,4 +26,11 @@ public function that(Expression $expression): AndThatShouldParser
 
         return new AndThatShould($this->ruleBuilder);
     }
+
+    public function exclude(string ...$classesToBeExcluded): ThatParser
```
fain182, line 30: "All classes exclude Pippo that reside in one of these namespaces..."
doesn't sound like english... maybe we could substitute exclude with "except" or "a part of"?

fain182:
A part from a naming detail, it seems a great solution and simple to explain... :top:

src/Rules/ArchRule.php
micheleorselli: yup, it would be nice to support wildcard :-) 

I guess the check can stay here

src/Rules/DSL/ThatParser.php
```diff
@@ -7,5 +7,7 @@
 
 interface ThatParser
 {
+    public function exclude(string ...$classesToBeExcluded): self;
```
micheleorselli, line 10: Should we define a new interface for the DSL? Placing that method here allow also this call chain

```
allClasses()
->exclude('Banana')
->that(ResideInaNamespace('Fruit')
->exclude('Mango')
->that(ImplementInterface('HasVitaminC'))
...
```

To me such syntax is misleading, since the call to `->exclude('Mango')` is ambiguous: does it exclude classes from the whole class set or only from the first `->that(ResideInaNamespace('Fruit')` invocation?

For this reason I think ->exclude should be callable right after allClasses() and in no other places

tests/E2E/Cli/phparkitect.php
```diff
@@ -13,11 +13,13 @@
     $mvc_class_set = ClassSet::fromDir(__DIR__.'/../_fixtures/mvc');
 
     $rule_1 = Rule::allClasses()
+        ->exclude('App\Controller\BaseController')
```
micheleorselli, line 16: If we want to keep consistency and have rules readable as english sentences `excluded` or `except` sound better to me, what do you think?

src/Rules/AllClasses.php
AlessandroMinoccheri: I think `except` is better.

tests/E2E/Cli/phparkitect.php
AlessandroMinoccheri: `except` seems better in my opinion, I agree.

src/Rules/AllClasses.php
```diff
@@ -11,9 +11,12 @@ class AllClasses implements ThatParser
 {
     /** @var RuleBuilder */
     protected $ruleBuilder;
+    /** @var array */
+    private $classesToBeExcluded;
```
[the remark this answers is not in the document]
AlessandroMinoccheri: No, I remove those lines.

src/Rules/ArchRule.php
AlessandroMinoccheri: Ok I added a new method to support wildcard here.

README.md
AlessandroMinoccheri: I added some lines

src/Rules/DSL/ThatParser.php
AlessandroMinoccheri: You can't use this syntax because `that` returns `AndThatShouldParser` interface.
So you can't use `except` after `that`

fain182 (approved):

micheleorselli (approved):

PR #225, opened 2022-01-06
implements have name start with and ends with class expressions

fokosun:
### What does this PR do?

Implements two new class expressions namely:  HaveNameEndingWith and  HaveNameStartingWith

### HaveNameEndingWith
A class expression that evaluates that the classname ends with a given string

### HaveNameStartingWith
A class expression that evaluates that the classname starts with a given string

### DIFF
--- a/src/Expression/ForClasses/HaveNameEndingWith.php
+++ b/src/Expression/ForClasses/HaveNameEndingWith.php
@@ -30,7 +30,9 @@ class HaveNameEndingWith implements Expression
     {
         $fqcn = FullyQualifiedClassName::fromString($theClass->getFQCN());
 
-        if (0 == strpos($fqcn->className(), $this->name)) {
+        $sub_str = substr($fqcn->className(), (-1 * strlen($this->name)));
+
+        if ($this->name !== $sub_str) {
             $violation = Violation::create(
                 $theClass->getFQCN(),
                 $this->describe($theClass)->toString()
--- a/src/Expression/ForClasses/HaveNameEndingWith.php
+++ b/src/Expression/ForClasses/HaveNameEndingWith.php
@@ -11,7 +11,7 @@ use Arkitect\Expression\PositiveDescription;
 use Arkitect\Rules\Violation;
 use Arkitect\Rules\Violations;
 
-class HaveNameEndsWith implements Expression
+class HaveNameEndingWith implements Expression
 {
     /** @var string */
     private $name;
@@ -30,7 +30,7 @@ class HaveNameEndsWith implements Expression
     {
         $fqcn = FullyQualifiedClassName::fromString($theClass->getFQCN());
 
-        if (!(0 != strpos($fqcn->className(), $this->name))) {
+        if (0 == strpos($fqcn->className(), $this->name)) {
             $violation = Violation::create(
                 $theClass->getFQCN(),
                 $this->describe($theClass)->toString()
--- a/src/Expression/ForClasses/HaveNameEndsWith.php
+++ b/src/Expression/ForClasses/HaveNameEndsWith.php
@@ -3,40 +3,40 @@ declare(strict_types=1);
 
 namespace Arkitect\Expression\ForClasses;
 
-use Arkitect\Rules\Violation;
-use Arkitect\Rules\Violations;
-use Arkitect\Expression\Expression;
-use Arkitect\Expression\Description;
 use Arkitect\Analyzer\ClassDescription;
 use Arkitect\Analyzer\FullyQualifiedClassName;
+use Arkitect\Expression\Description;
+use Arkitect\Expression\Expression;
 use Arkitect\Expression\PositiveDescription;
+use Arkitect\Rules\Violation;
+use Arkitect\Rules\Violations;
 
 class HaveNameEndsWith implements Expression
 {
-	/** @var string */
-	private $name;
+    /** @var string */
+    private $name;
 
-	public function __construct(string $name)
-	{
-		$this->name = $name;
-	}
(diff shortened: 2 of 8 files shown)


pfazzi:
Hello @fokosun, thanks for contributing :)

At the beginning of this project we have been debating about the minimum PHP version to support and we finally agreed to maintain backward compatibility with older PHP versions, since PHP 7.1. The reason is that we want to allow people anchored in ancient PHP versions to install and run the tool easily. 

Please, could you use `strpos`  instead of new functions in order to guarantee compatibility with older PHP versions? Thansk 🙂

fokosun:
> Hello @fokosun, thanks for contributing :)
> 
> At the beginning of this project we have been debating about the minimum PHP version to support and we finally agreed to maintain backward compatibility with older PHP versions, since PHP 7.1. The reason is that we want to allow people anchored in ancient PHP versions to install and run the tool easily.
> 
> Please, could you use `strpos` instead of new functions in order to guarantee compatibility with older PHP versions? Thansk 🙂

got it! working on that. Thanks for the feedback

pfazzi:
@fokosun you should comply to coding styles. Please run `make build` before to push ;)

src/Expression/ForClasses/HaveNameEndsWith.php
```diff
@@ -0,0 +1,42 @@
...
+
+use Arkitect\Analyzer\ClassDescription;
+use Arkitect\Analyzer\FullyQualifiedClassName;
+use Arkitect\Expression\Description;
+use Arkitect\Expression\Expression;
+use Arkitect\Expression\PositiveDescription;
+use Arkitect\Rules\Violation;
+use Arkitect\Rules\Violations;
+
+class HaveNameEndsWith implements Expression
```
pfazzi, line 14: English is not my mother language,  but let me ask the same.  Isn't `HaveNameEndingWith` more correct than `HaveNameEndsWith`?

src/Expression/ForClasses/HaveNameStartsWith.php
```diff
@@ -0,0 +1,42 @@
...
+    public function describe(ClassDescription $theClass): Description
+    {
+        return new PositiveDescription("should have a name that starts with {$this->name}");
+    }
+
+    public function evaluate(ClassDescription $theClass, Violations $violations): void
+    {
+        $fqcn = FullyQualifiedClassName::fromString($theClass->getFQCN());
+
+        if (!(0 === strpos($fqcn->className(), $this->name))) {
```
pfazzi, line 33: Can you simplify this condition? `0 !== strpos($fqcn->className(), $this->name)` seems more readable :)

tests/Unit/Expressions/ForClasses/HaveNameEndsWithTest.php
```diff
@@ -0,0 +1,35 @@
...
+    public function test_show_violation_when_class_name_does_not_end_with(): void
+    {
+        $expression = new HaveNameEndsWith('Suffix');
+
+        $badClass = ClassDescription::build('\App\BadNameClass')->get();
+
+        $violations = new Violations();
+        $expression->evaluate($badClass, $violations);
+        self::assertNotEquals(0, $violations->count());
+        $this->assertEquals('should have a name that ends with Suffix', $expression->describe($badClass)->toString());
```
pfazzi, line 33: It's better to be consistent with PHPUnit method calls. Use the static syntax in this case ;)

pfazzi (changes requested):

AlessandroMinoccheri:
@fokosun thanks for this PR!
I agree with @pfazzi comments and @fokosun, if you need some helps let me know.

tests/Unit/Expressions/ForClasses/HaveNameEndsWithTest.php
fokosun: Done

src/Expression/ForClasses/HaveNameStartsWith.php
fokosun: Done

src/Expression/ForClasses/HaveNameEndingWith.php
```diff
@@ -0,0 +1,44 @@
...
+    public function describe(ClassDescription $theClass): Description
+    {
+        return new PositiveDescription("should have a name that ends with {$this->name}");
+    }
+
+    public function evaluate(ClassDescription $theClass, Violations $violations): void
+    {
+        $fqcn = FullyQualifiedClassName::fromString($theClass->getFQCN());
+
+        $sub_str = substr($fqcn->className(), (-1 * strlen($this->name)));
```
pfazzi, line 33: Variable naming should be in camelCase :)

pfazzi (changes requested):

src/Expression/ForClasses/HaveNameEndingWith.php
fokosun: Apologies for the delay in submitting the requested change. I'll be pushing up the change soon. Thanks

diimpp:
Is there a reason to introduce new expressions, when `HaveNameMatching('*Controller')` already exists?

fokosun:
> Is there a reason to introduce new expressions, when `HaveNameMatching('*Controller')` already exists? If wildcard is not good enough, then perhaps generic regexp expression can be added

You have a valid point there! Apologies for the long delay in responding and working to address the issues on this PR. It looks like this PR is not really necessary then based on this comment. Is it okay to close the PR?

AlessandroMinoccheri:
Ok @fokosun I close this PR but feel free to contribute again :)

PR #227, opened 2022-01-07
adding phparkitect check for expression into ci

AlessandroMinoccheri:


phparkitect.php
```diff
@@ -0,0 +1,23 @@
...
+use Arkitect\Rules\Rule;
+
+return static function (Config $config): void {
+    $classSet = ClassSet::fromDir(__DIR__.'/src/Expression/ForClasses');
+
+    $rules = [];
+
+    $rules[] = Rule::allClasses()
+        ->that(new ResideInOneOfTheseNamespaces('Arkitect\Expression\ForClasses'))
+        ->should(new Implement('Arkitect\Expression\Expression'))
```
pfazzi, line 18: Can we use the more safe `Expression::class`?

micheleorselli:
That's a great addition, not sure if we should use the stub file to run the checks because that file is part of the stuff being tested. What about adding phparkitect as dev dependency and using it? What do you think?

fain182 (approved):

AlessandroMinoccheri:
@micheleorselli yes, I can try.
A good solution should be to use our github actions instead of using the command directly.

Issue #224, opened 2021-12-23
add violations count


ricfio:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Could a violations count be added in the phparkitect output?
It seems to me that at the moment it is not there.

micheleorselli:
that's probably a good addition, makes sense to me

AlessandroMinoccheri:
@ricfio, do you want to try to add this feature?

ricfio:
sure, it's very simple.
see https://github.com/phparkitect/arkitect/pull/235

AlessandroMinoccheri:
closed by #237

PR #235, opened 2022-01-21
add violations count on cli output (#224)

ricfio:


AlessandroMinoccheri:
Why the PR is in draft status? What does it miss @ricfio?

ricfio:
> Why the PR is in draft status? What does it miss @ricfio?

Because I have a pending question about the row 146 in `src/CLI/Check.php`.
Can you see my related comment on the 'Files changed' page?

image

In truth it is not a comment... I'm not sure this is the right way...
I have added a 'code review' request for my self?

AlessandroMinoccheri:
@ricfio I can't see your comment, but in my opinion, you can try to write the feature, and after we can discuss it together :)

src/CLI/Check.php
```diff
@@ -143,8 +143,8 @@ private function getConfigFilename(InputInterface $input)
 
     private function printViolations(Violations $violations, OutputInterface $output): void
     {
-        $output->writeln('<error>ERRORS!</error>');
         $output->writeln(sprintf('%s', $violations->toString()));
```
AlessandroMinoccheri, line 146: In my opinion, we can keep the `ERRORS` line when we start printing all the violations

AlessandroMinoccheri (changes requested):

src/CLI/Check.php
ricfio: I think the `ERRORS` string is superfluous and redundant at this point, are you sure you want hold it?
In this case, do you prefer rollback also the tests (modified with my last commit) based on `ERRORS` string?

AlessandroMinoccheri: We can try to remove it for now and understand later if it was necessary or more clear with it @ricfio

ricfio:
Ok, I have removed the last commit on this branch.
The new PR is here: https://github.com/phparkitect/arkitect/pull/237

PR #239, opened 2022-02-10
WIP    parsing dependencies recursevily

AlessandroMinoccheri:
In this PR I changed how we parse files with their dependencies.
If we find a dependency, now we check also inside it other dependencies recursively.

Now the collection of ClassDescription has the responsibility to check rules because we need to get all the dependencies of the class to be checked.


src/Analyzer/Parser.php
```diff
@@ -5,9 +5,11 @@
 
 interface Parser
 {
-    public function parse(string $fileContent, string $filename): void;
+    public function parse(string $fileContent, string $filename): array;
```
AlessandroMinoccheri, line 8: Now parse return an array of classDescriptions

src/Analyzer/FileParser.php
```diff
@@ -25,61 +25,169 @@ class FileParser implements Parser
...
+            $this->fileVisitor->clearParsedClassDescriptions();
+
+            if ($this->classDescriptionsParsed->exists($dependency->getFQCN()->toString())) {
+                continue;
+            }
+
+            try {
+                $classReflection = new \ReflectionClass($dependency->getFQCN()->toString());
+            } catch (\Exception $e) {
+                if (!$this->isAlreadyInErrors($dependency->getFQCN()->toString(), $e->getMessage())) {
```
AlessandroMinoccheri, line 114: is it correct to add it into errors?

src/Analyzer/ClassDescriptionCollection.php
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        $extends[] = $extendsFoundFQCN;
+        return $this->getExtends($extendsFoundFQCN->toString(), $extends);
+    }
+
+    public function getInterfaces(string $classDescriptionFQCN, array $interfaces = []): array
+    {
+        if (!array_key_exists($classDescriptionFQCN, $this->classDescriptions)) {
+            throw new \Exception('Class not found: ' . $classDescriptionFQCN . ' into collection');
```
AlessandroMinoccheri, line 81: is it correct to throw an exception here?

src/Analyzer/ClassDescriptionCollection.php
```diff
@@ -0,0 +1,116 @@
...
+
+    public function getClassDescriptions(): array
+    {
+        return $this->classDescriptions;
+    }
+
+    public function getDependencies(string $classDescriptionFQCN, array $dependencies = []): array
+    {
+        if (!array_key_exists($classDescriptionFQCN, $this->classDescriptions)) {
+            throw new \Exception('Class not found: ' . $classDescriptionFQCN . ' into collection');
```
AlessandroMinoccheri, line 30: is it correct to throw an exception here?

src/Analyzer/ClassDescriptionCollection.php
```diff
@@ -0,0 +1,116 @@
...
+
+    public function exists(string $classDescriptionFQCN): bool
+    {
+        return array_key_exists($classDescriptionFQCN, $this->classDescriptions);
+    }
+
+    public function get(string $FQCN): ClassDescription
+    {
+        if (!$this->exists($FQCN)) {
+            throw new \Exception('Class not found: ' . $FQCN . ' into collection');
```
AlessandroMinoccheri, line 110: is it correct to throw an exception here?

AlessandroMinoccheri:
Tests aren't green. because in FileVisitorTest we have something like this:

```
    $code = <<< 'EOF'
<?php

namespace MyNamespace\MyClasses;

use AnotherNamespace\Baz;

class Foo
{
    public function foo()
    {
        $bar = new Bar();
        $baz = new Baz();
    }
}
EOF;
```

With the new parser logic, we check also `Baz` file with the reflection and we can't find it in this code.
We need to understand if it is correct to generate an error if we can't use Reflection, or if we need to change all tests like this one with another strategy.

pfazzi:
La chiudiamo questa? È aperta da tempo immemore... :D

Issue #242, opened 2022-02-25
phparkitect.phar does not work on PHP 7.3


LuigiCardamone:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | yes/no
| Library Version | 0.2.15
| PHP version     | 7.3.33

#### Summary

phparkitect.phar does not work on PHP 7.3

#### Current behavior

I get this error:
```
Your Composer dependencies require a PHP version ">= 7.4.0". You are running 7.3.33-1+ubuntu18.04.1+deb.sury.org+1.

PHP Fatal error:  Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 7.4.0". You are running 7.3.33-1+ubuntu18.04.1+deb.sury.org+1. in phar:///var/www/app/phparkitect.phar/vendor/composer/platform_check.php on line 24
```

#### How to reproduce

On Ubuntu 18 with PHP 7.3 the error appears following the basic setup instructions:
```
wget https://github.com/phparkitect/arkitect/releases/latest/download/phparkitect.phar
chmod +x phparkitect.phar
./phparkitect.phar check

```

Any idea?
(I know that PHP 7.3 is in EOL, but it is not up on me :-) )

AlessandroMinoccheri:
Hi @LuigiCardamone 
I am not sure if it can work or not, but can you try to specify the target PHP version In this way?

```
./phparkitect.phar check --target-php-version=7.3
```

Let me know.

pfazzi:
Ciao @LuigiCardamone :)

LuigiCardamone:
Hi @AlessandroMinoccheri, nope, same error.
I will try in a project with PHP 7.4, thanks anyway.

@pfazzi Ciao! Congratulations for this very interesting project!

AlessandroMinoccheri:
@LuigiCardamone can you install it via `composer`?

LuigiCardamone:
Unfortunately no: several dependencies conflicts on our main (and huge) project. I will try on a test project

AlessandroMinoccheri:
@LuigiCardamone, have you tried phparkitect in a project with PHP 7.4?
Can you also share more information about your problem with the PHP7.3 project? For example composer constraints or something relevant to reproduce the problem.

pfazzi:
We could try to reproduce the issue with a docker container, what do you think guys?

AlessandroMinoccheri:
Yes, it's important to reproduce a similar error so with the right constraints.

[...]

PR #245, opened 2022-03-05
changed php version during phar creation

AlessandroMinoccheri:
I changed the PHP version from `7.4` to `7.1` to solve this issue: #242 
In my opinion, this can create future problems, I don't know at the moment if the option `target-php-version` is enough or we have to create a phar version for each PHP version.

### DIFF
--- a/.github/workflows/build-phar.yml
+++ b/.github/workflows/build-phar.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Install PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 7.4
+          php-version: 7.1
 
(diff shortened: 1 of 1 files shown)


fain182:
Do you think it might be worth creating some tests?

AlessandroMinoccheri:
Yes @fain182, but you need to create a phar file and try it with different containers with different PHP versions.

micheleorselli:
Makes sense to me as we should use the lowest PHP version we support in composer.json

IIRC that was the case at the beginning, did we change it for a reason? 🤔

micheleorselli (approved):
Looks good to me then :-)

PR #249, opened 2022-05-21
Increase the psalm error level from 3 to 2

AlessandroMinoccheri:


micheleorselli:
it seems the build problem is related to composer, as it complains about unsupported PHP version

AlessandroMinoccheri:
The same problem is in the other PR too:
https://github.com/phparkitect/arkitect/pull/252

We can try to use for example `version: 1` instead of `version: 2` @micheleorselli

AlessandroMinoccheri:
Yeah, it was that now it's all green.
Do you think that it's better to solve the problem in another little PR moving last commit from here @micheleorselli ?

micheleorselli:
let's stick with version 2. composer 1 is unsupported and slow as hell. Version [2.2.12](https://github.com/composer/composer/releases/tag/2.2.12) should be compatible with our PHP version

AlessandroMinoccheri:
@micheleorselli I tried with composer 2 and that specific version without success.

micheleorselli (approved):

PR #253, opened 2022-05-25
tries to fix composer issues

micheleorselli:
@AlessandroMinoccheri this should fix the composer issue


AlessandroMinoccheri (approved):
Amazing 👍

Issue #229, opened 2022-01-09
New Expression BeAnnotatedWith


pfazzi:
### Feature Request

Add `BeAnnotatedWith` Expression.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes/no
| BC Break    | yes/no

#### Summary

Use case:

```php
yield Rule::allClasses()
        ->that(new Implement(App\Shared\Domain\Event\Event::class))
        ->should(new BeAnnotatedWith('@psalm-immutable'))
        ->because('we want to enforce immutability');
```

ricfio:
As class naming I suggest `HaveAnnotationMatching` or `HaveAnnotations` (this last one is my preferred) instead of `BeAnnotatedWith`.

AlessandroMinoccheri:
Closed by #255

PR #255, opened 2022-05-27
adding HaveAnnotation rule

AlessandroMinoccheri:
This PR should solve #229

### DIFF
--- a/src/Analyzer/ClassDescription.php
+++ b/src/Analyzer/ClassDescription.php
@@ -27,13 +27,17 @@ class ClassDescription
     /** @var bool */
     private $abstract;
 
+    /** @var string */
+    private $docBlock;
+
     public function __construct(
         FullyQualifiedClassName $FQCN,
         array $dependencies,
         array $interfaces,
         ?FullyQualifiedClassName $extends,
         bool $final,
-        bool $abstract
+        bool $abstract,
+        string $docBlock = ''
     ) {
         $this->FQCN = $FQCN;
         $this->dependencies = $dependencies;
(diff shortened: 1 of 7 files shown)


src/Analyzer/ClassDescription.php
```diff
@@ -141,4 +146,18 @@ public function isAbstract(): bool
     {
         return $this->abstract;
     }
+
+    public function getAnnotation(): string
```
fain182, line 150: this is a container of multiple annotations? maybe this should be plural to be more coherent with hasAnnotation?

AlessandroMinoccheri: yes, I agree, I fix it immediately.

fain182 (approved):
:+1:

micheleorselli (commented):
I do like the idea behind this PR, but I am not sure I fully understood the scope. We are saying we want to check if a cIass has a specific annotation? 

If is that so, I am wondering if the name `Annotation` could be misleading since we are checking if a DocBlock contains a string, so for instance if we have this snippet of code

```php

/**
* This is just a comment
*/
class Banana {}
```

the rule `new HaveAnnotation('just a')` will be matched. Usually, when we say "annotation" we think about something like this

```php
/**
* @annotation 
*/
```

AlessandroMinoccheri:
Good point @micheleorselli could it be better `new HaveDocBlock('just a')` or is there a better name?

micheleorselli:
> Good point @micheleorselli could it be better `new HaveDocBlock('just a')` or is there a better name?

DocBlockContains?
HaveDocBlockContaining?

micheleorselli (approved):

Issue #258, opened 2022-06-04
how to use `because()`?


hgraca:
|    Q            |   A
|---------------- | ------
| Library version | 0.2.18
| PHP version     | 7.2

### Support Question

Hi, I'm trying to figure out how to show the `because()` when rules fail, but I cant figure out how.

So, when we have a rule like 

```php
$rule_1 = Rule::allClasses()
        ->that(new ResideInOneOfTheseNamespaces('App\Infrastructure\Controller'))
        ->should(new HaveNameMatching('*Controller'))
        ->because('we want to uniform controller name.');
```
When this rule fails, the error is 

`should have a name that matches *Controller`

but it would be great if it would show 

`should have a name that matches *Controller because we want to uniform controller name.`.

This would be helpful to explain to others what is the reasoning behind the rule.

AlessandroMinoccheri:
Hi @hgraca at the moment the `because` is used only when you write the rule to specify why you added it.
But I think it can be helpful also in the error message.

Do you want to try to create a PR to solve the problem?

hgraca:
This is awesome ppl!!!
I'm sorry, I missed your first message.

Thanks a lot, and keep it up!

I have created a few rules for the project I'm working on, I will submit MRs.

AlessandroMinoccheri:
Thanks to you @hgraca :)

PR #259, opened 2022-06-06
added because reason to error messages

AlessandroMinoccheri:
In this PR I added the because string to error messages so if previously we printed this violation:

```
App\Controller\Foo violates rules  should implement ContainerAwareInterface
```

Now the violation error will be:

```
App\Controller\Foo violates rules  should implement ContainerAwareInterface  because all controllers should be container aware
```

This PR solve this issue: #258


src/Expression/ForClasses/DependsOnlyOnTheseNamespaces.php
```diff
@@ -21,14 +21,18 @@ public function __construct(string ...$namespace)
         $this->namespaces = $namespace;
     }
 
-    public function describe(ClassDescription $theClass): Description
+    public function describe(ClassDescription $theClass, string $because): Description
     {
         $desc = implode(', ', $this->namespaces);
 
-        return new PositiveDescription("should depend only on classes in one of these namespaces: $desc");
+        if ('' !== $because) {
```
fain182, line 28: Maybe we can add the because as a second argument of the Positive Description constructor, so that we don't need to repeat the if on the empty because everywhere?

micheleorselli: +1 for @fain182 suggestion

AlessandroMinoccheri: @fain182so are you suggesting to change the PositiveDescription like this?

```
class PositiveDescription implements Description
{
    /** @var string */
    private $description;

    public function __construct(string $description, string $because)
    {
        $this->description = $description;
        if ('' !== $because) {
            $this->description .= ' because' . $because
       }
    }
```

fain182: Exactly... 
(maybe we could refactor a little more around PositiveDescription and NegativeDescription but we will think about it in another PR)

AlessandroMinoccheri: @fain182 I move the because concatenation logic into the PositiveDescription

fain182 (approved):

Issue #256, opened 2022-05-31
Add stop-on-failure command option


AlessandroMinoccheri:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

It could be interesting to add a `stop-on-failure` to avoid waiting for the end of the entire process.
What do you think?

AlessandroMinoccheri:
Closed by #262

PR #262, opened 2022-06-12
adds stop on failure option using exceptions

fain182:
This PR solve #256

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,4 +175,4 @@ phparkitect check --config=/project/yourConfigFile.php
 ```
 * `--target-php-version`: with this parameter, you can specify which PHP version should use the parser. This can be useful to debug problems and to understand if there are problems with a different PHP version.
 Supported PHP versions are: 7.1, 7.2, 7.3, 7.4, 8.0, 8.1
- * `-stop-on-failure` : with this option if there is a violation, the process will end immediately
+ * `--stop-on-failure` : with this option the process will end immediately after the first violation.
--- a/README.md
+++ b/README.md
@@ -175,3 +175,4 @@ phparkitect check --config=/project/yourConfigFile.php
 ```
 * `--target-php-version`: with this parameter, you can specify which PHP version should use the parser. This can be useful to debug problems and to understand if there are problems with a different PHP version.
 Supported PHP versions are: 7.1, 7.2, 7.3, 7.4, 8.0, 8.1
+ * `-stop-on-failure` : with this option if there is a violation, the process will end immediately
--- a/src/CLI/Check.php
+++ b/src/CLI/Check.php
@@ -54,7 +54,7 @@ class Check extends Command
             ->addOption(
                 self::STOP_ON_FAILURE_PARAM,
                 's',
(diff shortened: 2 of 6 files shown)


src/CLI/Check.php
```diff
@@ -72,8 +86,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
             $this->readRules($config, $rulesFilename);
 
-            $runner = new Runner();
-            $runner->run($config, $progress, $targetPhpVersion);
+            $runner = new Runner($stopOnFailure);
+            try {
+                $runner->run($config, $progress, $targetPhpVersion);
+            } catch (FailOnFirstViolationException $e) {
```
fain182, line 92: this empty catch is the worst part in my opinion...

AlessandroMinoccheri: Uhm, yes but this solution is much better and clean IMHO.

AlessandroMinoccheri (commented):
This solution is much better and clean in my opinion :)

AlessandroMinoccheri (approved):

src/CLI/Check.php
```diff
@@ -46,6 +50,12 @@ protected function configure(): void
                 't',
                 InputOption::VALUE_OPTIONAL,
                 'Target php version to use for parsing'
+            )
+            ->addOption(
+                self::STOP_ON_FAILURE_PARAM,
+                's',
+                InputOption::VALUE_OPTIONAL,
+                'Stop on failure'
```
AlessandroMinoccheri, line 58: We can try to add the default value and manage it like in this link,
https://symfony.com/doc/current/console/input.html#options-with-optional-arguments

What do you think @fain182 ? The Symfony example represents our case, the behavior of the optional parameter.

fain182: I studied it a little, and I think that this is not a VALUE_OPTIONAL, but a VALUE_NONE, because we will not pass any argument... and with VALUE_NONE I can't set the default, but anyway is false, like you correctly thought...

AlessandroMinoccheri: Great!

AlessandroMinoccheri (changes requested):
Can you please edit the README file with this new option?

fain182:
README fixed :slightly_smiling_face:

README.md
```diff
@@ -175,3 +175,4 @@ phparkitect check --config=/project/yourConfigFile.php
 ```
 * `--target-php-version`: with this parameter, you can specify which PHP version should use the parser. This can be useful to debug problems and to understand if there are problems with a different PHP version.
 Supported PHP versions are: 7.1, 7.2, 7.3, 7.4, 8.0, 8.1
+ * `-stop-on-failure` : with this option if there is a violation, the process will end immediately
```
AlessandroMinoccheri, line 178: I tried the option and now we can use `--stop-on-failure` and not `-stop-on-failure` 
Am I right?

AlessandroMinoccheri (changes requested):

AlessandroMinoccheri (approved):

PR #263, opened 2022-06-13
feat: Add `HaveAttribute` class expression

frankverhoeven:
Hi there!

This change will add the ability to check for the existence of an attribute on a class. The addition of this ability won't break existing functionality on PHP < 8.0, but because attributes only exist on PHP >= 8.0, using this new expression does require the latter. Perhaps this limitation should be documented explicitly if you are willing to accept the change.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Currently, you can check if a class:
  - is not abstract
  - doc block contains a string
  - doc block not contains a string
+ - has an attribute
 
 You can also define components and ensure that a component:
 - should not depend on any component
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Currently, you can check if a class:
  - is not abstract
  - doc block contains a string
  - doc block not contains a string
- - has an attribute
+ - has an attribute (requires PHP >= 8.0)
 
 You can also define components and ensure that a component:
 - should not depend on any component
--- a/src/Analyzer/ClassDescription.php
+++ b/src/Analyzer/ClassDescription.php
@@ -8,17 +8,16 @@ class ClassDescription
     /** @var FullyQualifiedClassName */
     private $FQCN;
 
-    /** @var array */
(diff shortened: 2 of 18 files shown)


README.md
```diff
@@ -32,6 +32,7 @@ Currently, you can check if a class:
  - is not abstract
  - doc block contains a string
  - doc block not contains a string
+ - has an attribute
```
AlessandroMinoccheri, line 35: I think that it could be useful to specify that this rule can work only for PHP > 8, what do you think?

fain182:
Hi Frank, thank you for your contribution! :partying_face:  

To understand better your needs, can you tell me the use case in which you want to check the presence of an attribute in a group of classes?
(Feel free to attach an example of architectural test, if you already have one with this rule :slightly_smiling_face: )

frankverhoeven:
I think the best example is the [functional test](https://github.com/phparkitect/arkitect/pull/263/files#diff-68e8e82d97df4602ab81661cff773ae44a8be832fdae132261efd6caa9655a8f) that is included.

So for example, Symfony has an `AsController` attribute: https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/HttpKernel/Attribute/AsController.php. Instead of saying, `every class in the namespace '*Controller*' must have a name ending in 'Controller'`, I could more specifically say `every class that has the 'AsController' attribute must have a name ending in 'Controller'`.

This in turn would allow me to place other related classes, eg a normaliser that is solely used with a certain controller, under the same namespace, without forcing the `Controller` suffix on that normaliser. Thus, group classes by concern rather than grouping them by type, while still enforcing certain conventions.

fain182:
ook, great idea! :slightly_smiling_face:

AlessandroMinoccheri (approved):
LGTM 👍 
Thanks for your contribution!

Issue #254, opened 2022-05-27
Adding more examples


AlessandroMinoccheri:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

We can improve the README file by adding an example for each rule.

PR #266, opened 2022-07-02
Adding specific examples for each rule

AlessandroMinoccheri:
This PR should solve #254 

I removed the list of rules with specific examples.
We should add a description for each rule? Or is it clear?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,24 +15,169 @@ Rule::allClasses()
 ## What kind of rules can I enforce with Arkitect
 
 Currently, you can check if a class:
- - depends on a namespace
- - extends another class
- - not extends another class
- - have a name matching a pattern
- - not have a name matching a pattern
- - implements an interface
- - not implements an interface
- - depends on a namespace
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,24 +15,169 @@ Rule::allClasses()
...
+    ->because('we want protect our domain except for Ramsey\Uuid');
+```
+
+### Not have a name matching a pattern
+
+```
+$rules = Rule::allClasses()
+    ->that(new ResideInOneOfTheseNamespaces('App'))
+    ->should(new NotHaveNameMatching('*Manager'))
+    ->because('we want to avoid using Manager for our classes');
```
fain182, line 160: Maybe something more like:
"*Manager is too vague in naming classes"

fain182 (approved):
Much better than before.

README.md
AlessandroMinoccheri: I agree, much better.

PR #268, opened 2022-07-09
added prophecy as a dependency to avoid warning into our tests

AlessandroMinoccheri:
In this PR I added the dependency phpspec/prophecy-phpunit to use the ProphecyTrait in our tests and avoid warning when we run them.


AlessandroMinoccheri:
We can't add ProphecyTrait because it requires PHP > 7.3

PR #269, opened 2022-07-09
test: add test for const outside namespace

ricfio:
I have added a FAIL test to detect const outside namespace as dependency.


AlessandroMinoccheri:
Thanks for this PR @ricfio, it was very useful!

PR #276, opened 2022-07-22
Parsing enum file

### LINKED ISSUE #275: Parse Enum
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

We should add the possibility to parse an Enum file.

AlessandroMinoccheri:
This PR should fix #275

### DIFF
--- a/src/Analyzer/FileVisitor.php
+++ b/src/Analyzer/FileVisitor.php
@@ -126,6 +126,13 @@ class FileVisitor extends NodeVisitorAbstract
                 ->addDependency(new ClassDependency($node->class->toString(), $node->getLine()));
         }
 
+        if ($node instanceof Node\Stmt\Enum_ && null !== $node->namespacedName) {
+            /** @psalm-suppress UndefinedPropertyFetch */
+            $this->classDescriptionBuilder = ClassDescriptionBuilder::create(
+                $node->namespacedName->toCodeString()
+            );
+        }
+
         /**
          * matches parameters dependency in functions and method definitions like
          * public function __construct(Symfony\Component\HttpFoundation\Request $request).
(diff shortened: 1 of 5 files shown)


tests/Unit/Analyzer/FileVisitorTest.php
```diff
@@ -419,4 +420,32 @@ public function getStatic(): self
...
+        /** @var FileParser $fp */
+        $fp = FileParserFactory::createFileParser(TargetPhpVersion::create('8.1'));
+        $fp->parse($code, 'relativePathName');
+
+        $cd = $fp->getClassDescriptions();
+
+        $violations = new Violations();
+
+        $notHaveDependencyOutsideNamespace = new Implement('MyInterface');
+        $notHaveDependencyOutsideNamespace->evaluate($cd[0], $violations, 'we want to add this rule for our software');
```
micheleorselli, line 447: I am not sure I got the goal of this test

src/Analyzer/FileVisitor.php
```diff
@@ -154,6 +161,12 @@ public function leaveNode(Node $node): void
 
             $this->classDescriptions[] = $classDescription;
         }
+
+        if ($node instanceof Node\Stmt\Enum_ && null !== $this->classDescriptionBuilder) {
```
micheleorselli, line 165: Here we treat enum as a class which is fine for now. I am wondering if we should add a method like isEnum or similar

AlessandroMinoccheri: What is the benefit to add the method `isEnum` now?
I think it could be fine.

tests/Unit/Analyzer/FileVisitorTest.php
AlessandroMinoccheri: It's a simple rule @micheleorselli to check if we are parsing or not the enum class.
Without the code inside the `FileVisitor` the enum class is not parsed and this test it's red.
We can try to find a different test to be more explicit. Any ideas?

Issue #277, opened 2022-08-02
Only depends on components


alessandro-candon:
###  onlyDependOnComponents

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         |  no
| BC Break    | no

#### Summary

Some classes should ONLY depends on others components.

For example, i have this situation to validate...

<img width="296" alt="image" src="https://user-images.githubusercontent.com/44228481/182334438-e3a23cf6-b38e-4fd8-a0b0-bb030e964399.png">

```
Filters
   Appliers : the filtering logic
      StockModel
      StockSku
   Shared : all logic that allow filters work
```
StockModel &  StockSku should only have 2 files that
1. Has a suffix of   _Applier_ and _DTO_ 
2. Depends **ONLY** to the Shared classes

Is it possible to do with the rules already written?
If not, do you think is a possible new good feature?

Thank you so much!

AlessandroMinoccheri:
Hi @alessandro-candon,
thanks for this issue, it's very interesting what you want to achieve.

1. At the moment we don't have the rule to specify if it's inside a class there is a specific number of files or more than one suffix
2. for this rule I usually use `DependsOnlyOnTheseNamespaces` but I think that in your example we need more than that right?

I think that we need to implement something, we can try to write an example of a config file to achieve the rules that you want @alessandro-candon 

After that, we can try to implement what is necessary.

@fain182 @micheleorselli @pfazzi any different suggestions?

alessandro-candon:
I think, as you say, it's complex than I thought.

I try to write a rule that can explain what we need :

    $try = Architecture::withComponents()
        ->component('FilterApplier')->definedBy('App\Filters\Appliers\**\**\*FilterApplier')
        ->component('FilterDto')->definedBy('App\Filters\Appliers\**\**\*FilterDto')
        ->component('FilterShared')->definedBy('App\Filters\Shared\*')
        ->where('FilterApplier')->onlyDependOnComponents('FilterDto', 'FilterShared')
        ->rules();

So, maybe what we need is 
- support `definedBy` with regex
- introduce `onlyDependOnComponents` to block every type of dependency with other classes...

An other possibility is to introduce `AND` `OR` operator in this way you define some rules that exists at the same time...

fain182:
It could be a good idea...

[...]

PR #278, opened 2022-08-03
added add-shouldOnlyDependOnComponents

AlessandroMinoccheri:
In this PR I added the possibility to filter components with `shouldOnlyDependOnComponents` function that generates a rule: `DependsOnlyOnTheseNamespaces`

This can fix some problems about issue #277

### DIFF
--- a/src/RuleBuilders/Architecture/Architecture.php
+++ b/src/RuleBuilders/Architecture/Architecture.php
@@ -17,14 +17,14 @@ class Architecture implements Component, DefinedBy, Where, MayDependOnComponents
     /** @var array<string, string[]> */
     private $allowedDependencies;
     /** @var array<string, string[]> */
-    private $allowedSpecificDependencies;
+    private $componentDependsOnlyOnTheseNamespaces;
 
     private function __construct()
     {
         $this->componentName = '';
         $this->componentSelectors = [];
         $this->allowedDependencies = [];
-        $this->allowedSpecificDependencies = [];
+        $this->componentDependsOnlyOnTheseNamespaces = [];
     }
 
     public static function withComponents(): Component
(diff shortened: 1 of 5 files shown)


src/RuleBuilders/Architecture/Architecture.php
```diff
@@ -3,24 +3,28 @@
...
+class Architecture implements Component, DefinedBy, Where, MayDependOnComponents, MayDependOnAnyComponent, ShouldNotDependOnAnyComponent, ShouldOnlyDependOnComponents, Rules
 {
     /** @var string */
     private $componentName;
     /** @var array<string, string> */
     private $componentSelectors;
     /** @var array<string, string[]> */
     private $allowedDependencies;
+    /** @var array<string, string[]> */
+    private $allowedSpecificDependencies;
```
pfazzi, line 20: Cosa cambia tra le `allowedDependencies` e le `allowedSpecificDependencies`?

AlessandroMinoccheri: `allowedDependencies ` is used for creating the rule `NotDependsOnTheseNamespaces`.
`allowedSpecificDependencies` is used for creating the rule `DependsOnlyOnTheseNamespaces`.

I don't like the design now but I would like to iterate on it later. 
What do you think @pfazzi ?

pfazzi: Ok, I think the implementation is a bit misleading ATM. We should iterate over it, but it could lead to a breaking change. 
Anyway in this step I would rename this variable to something more self-explaining, like `componentDependsOnlyOnTheseNamespaces`.

pfazzi (approved):

src/RuleBuilders/Architecture/Architecture.php
AlessandroMinoccheri: I agree with you @pfazzi. 
I renamed the variable with your suggestion :)

PR #281, opened 2022-08-16
Fix #126 improve error messages including classname when possible

### LINKED ISSUE #126: improve error messages including class name of the dependency when possibile

annervisser:
- Added ViolationDescription, as input for a Violation
- Added a description to the ViolationDescription
  - Only for Expressions where it made sense

### DIFF
--- a/src/Expression/ForClasses/DependsOnlyOnTheseNamespaces.php
+++ b/src/Expression/ForClasses/DependsOnlyOnTheseNamespaces.php
@@ -9,6 +9,7 @@ use Arkitect\Analyzer\ClassDescription;
 use Arkitect\Expression\Description;
 use Arkitect\Expression\Expression;
 use Arkitect\Rules\Violation;
+use Arkitect\Rules\ViolationMessage;
 use Arkitect\Rules\Violations;
 
 class DependsOnlyOnTheseNamespaces implements Expression
@@ -44,7 +45,10 @@ class DependsOnlyOnTheseNamespaces implements Expression
             if (!$dependency->matchesOneOf(...$this->namespaces)) {
                 $violation = Violation::createWithErrorLine(
                     $theClass->getFQCN(),
-                    $this->describe($theClass, $because)->toString(),
+                    ViolationMessage::withDescription(
+                        $this->describe($theClass, $because),
+                        "depends on {$dependency->getFQCN()->toString()}"
+                    ),
                     $dependency->getLine()
                 );
 
--- a/src/Expression/ForClasses/DocBlockContains.php
+++ b/src/Expression/ForClasses/DocBlockContains.php
@@ -8,6 +8,7 @@ use Arkitect\Analyzer\ClassDescription;
 use Arkitect\Expression\Description;
 use Arkitect\Expression\Expression;
 use Arkitect\Rules\Violation;
+use Arkitect\Rules\ViolationMessage;
 use Arkitect\Rules\Violations;
 
 class DocBlockContains implements Expression
@@ -30,7 +31,7 @@ class DocBlockContains implements Expression
         if (!$theClass->containsDocBlock($this->docBlock)) {
             $violation = Violation::create(
                 $theClass->getFQCN(),
-                $this->describe($theClass, $because)->toString()
+                ViolationMessage::selfExplanatory($this->describe($theClass, $because))
             );
             $violations->add($violation);
         }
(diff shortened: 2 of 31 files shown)


annervisser:
I was trying out Arkitect, and it looks like a great tool :smile:. The php config is great, so much better than Yaml.

I bumped into these messages not being quite as descriptive as I'd like. So I picked up this issue.

Let me know if this is something you'd like to merge and if anything needs to be changed :+1:

AlessandroMinoccheri:
Hi @annervisser, thanks for this PR.
Before starting the review, can you please fix the coding standards 😄?
Thanks.

annervisser:
@AlessandroMinoccheri done!

src/Expression/ExpressionDescription.php
```diff
@@ -4,7 +4,7 @@
 
 namespace Arkitect\Expression;
 
-class Description
+class ExpressionDescription
```
AlessandroMinoccheri, line 7: Why is it better this name for you instead of `Description`? 
Having already `Expression` in the namespace it sounds like a repetition or not?

src/Expression/ViolationDescription.php
```diff
@@ -0,0 +1,38 @@
...
+    /** @var string|null */
+    private $violation;
+
+    private function __construct(string $rule, ?string $violation)
+    {
+        $this->rule = $rule;
+        $this->violation = $violation;
+    }
+
+    public static function withDescription(ExpressionDescription $brokenRule, string $description): self
```
AlessandroMinoccheri, line 20: I think that we can improve the name of this class without using two times `Description`: in the class name and in the method name. What do you think?

src/Expression/ExpressionDescription.php
annervisser: I renamed it since otherwise there'd be a `Description` and a `ViolationDescription`

Maybe I should move `ViolationDescription` to be in `Arkitect\Rules`? (next to `Violation`) Then I could revert this to be Description without any confusion

src/Expression/ViolationDescription.php
annervisser: Agreed. How about ViolationMessage?

annervisser:
- reverted the rename of `Description` -> `ExpressionDescription`
- moved `ViolationDescription` from `Arkitect\Expression` -> `Arkitect\Rules` (to be next to `Violation`)
- renamed `ViolationDescription` to `ViolationMessage` to avoid duplicating the term description in classname and method

fain182:
Thank you @annervisser for your PR! I think that is absolutely necessary to explicit which dependency is raising the error.

I am not sure about the wording:

> App\Domain\Model violates **rules**
>   **depends on App\Services\UserService**, but should not depend on classes outside namespace App\Domain because we want protect our domain (on line 14) 

It seems that "depends on..." is part of the rule, but it's the opposite, it's the violation :thinking: 

Maybe we could write something like this? Other ideas?

> App\Domain\Model violates **rules**
>   should not depend on classes outside namespace App\Domain because we want protect our domain (it depends on App\Services\UserService on line 14)

src/Expression/ExpressionDescription.php
AlessandroMinoccheri: I think that we can do it in another PR after this one because it can have a different scope. At the moment it's good enough this name.

src/Expression/ViolationDescription.php
AlessandroMinoccheri: Sounds good to me

annervisser:
@fain182 I agree the wording is a bit weird right now.

I would be okay with your suggestion of putting it at the end with the line number

An alternative would be to change the wording to make it clear that they're violations (change `<fqcn> violates rules` to for example `<fqcn>  has <n> violations`
Maybe we could even group the messages per violation? like this: (though this is probably out of scope for this PR)
```
App\Domain\Model has 2 violations:
  rule: should not depend on classes outside namespace App\Domain because we want protect our domain
    - depends on App\Services\UserService (on line 14)
    - depends on App\Services\OtherService (on line 26)
```

As you said the current wording implies we're outputting rules, despite actually outputting violations. Currently this is already noticeable through the fact that a rule can be listed multiple times per file (when a rule has multiple violations)

What do you think?
My suggestion would be to change the wording to make it clear we're listing violations rather than rules, and follow up in a future PR to group violations per rule

fain182:
I agree with your idea @annervisser, we are showing violations, not rules right now  :+1:

annervisser:
@fain182 done in [`c9b3be0` (#281)](https://github.com/phparkitect/arkitect/pull/281/commits/c9b3be06789a32ff9ae80bd56207f01af6884384)

AlessandroMinoccheri (approved):
LGTM 👍

PR #283, opened 2022-08-17
added tests for Violations class

AlessandroMinoccheri:


tests/Unit/Rules/ViolationsTest.php
```diff
@@ -72,4 +72,36 @@ public function test_to_string(): void
 
         $this->assertEquals($expected, $this->violationStore->toString());
     }
+
+    public function test_get_iterable(): void
```
fain182, line 76: I suppose that the type checker already check this?

AlessandroMinoccheri: Yes, I was trying to explicit it well. I would like to be more specific, any suggestions @fain182 ?

AlessandroMinoccheri: @fain182 I changed the assertion with a more proper one, what do you think?

fain182: much better :+1:

fain182 (approved):

PR #285, opened 2022-08-18
fix bug when a wrong composer.json was used to read the version number

fain182:
There was an error in showing the version number when phparkitect is installed via composer.

You can reproduce the bug running https://github.com/phparkitect/arkitect-demo .

Schermata del 2022-08-18 11-22-11

Now it should work in all cases.

### DIFF
--- a/src/CLI/Version.php
+++ b/src/CLI/Version.php
@@ -6,7 +6,6 @@ namespace Arkitect\CLI;
 
 class Version
 {
-
     public static function get(): string
     {
         $pharPath = \Phar::running();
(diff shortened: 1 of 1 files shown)


micheleorselli (approved):

AlessandroMinoccheri (approved):
LGTM 👍 
It could be amazing to test more in this class, we can think about it in another PR.

fain182:
> LGTM +1 It could be amazing to test more in this class, we can think about it in another PR.

@AlessandroMinoccheri yes, it would be cool, I thought about it but the problem is that we should do some integration testing using a real phar and it's complicated.

PR #288, opened 2022-08-19
simplify init

fain182:
I am not sure about this one, the different includes seems redundant, without this code the CI is green...
Any additional test we can make to check if this is useful?

### DIFF
--- a/bin-stub/phparkitect
+++ b/bin-stub/phparkitect
@@ -12,7 +12,8 @@ if (isset($_composer_autoload_path)) {
 
 } else {
 // if $_composer_autoload_path is not set, we will try to find the autoload.php file
-// guessing the location of the phparkitect binary:
+// guessing the location of the phparkitect binary
+// ps: e2e tests are run from the bin-stub dir and needs that
     $possible_autoload_locations = [ 
         __DIR__ . '/../../autoload.php',        // inside vendor dir like "vendor/phparkitect/bin"
         __DIR__ . '/../vendor/autoload.php',    // first level project dir like "bin"
--- a/bin-stub/phparkitect
+++ b/bin-stub/phparkitect
@@ -4,7 +4,41 @@ declare(strict_types=1);
 
 use Arkitect\CLI\PhpArkitectApplication;
 
-require __DIR__ . '/../vendor/autoload.php';
+// $_composer_autoload_path is set by composer (available from composer 2.2)
+// https://getcomposer.org/doc/articles/vendor-binaries.md#finding-the-composer-autoloader-from-a-binary
+if (isset($_composer_autoload_path)) {
+    
+    define('PHPARKITECT_COMPOSER_INSTALL', $_composer_autoload_path);
+
+} else {
+// if $_composer_autoload_path is not set, we will try to find the autoload.php file
+// guessing the location of the phparkitect binary:
+    $possible_autoload_locations = [ 
+        __DIR__ . '/../../autoload.php',        // inside vendor dir like "vendor/phparkitect/bin"
+        __DIR__ . '/../vendor/autoload.php',    // first level project dir like "bin"
+        __DIR__ . '/vendor/autoload.php'        // project root
+    ];
+    
+    foreach ($possible_autoload_locations as $path) {
+        if (file_exists($path)) {
+            define('PHPARKITECT_COMPOSER_INSTALL', $path);
+            break;
+        }
+    }
+}
+
+if (!defined('PHPARKITECT_COMPOSER_INSTALL')) {
+    fwrite(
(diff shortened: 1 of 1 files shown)


AlessandroMinoccheri:
Did you try to generate the phar file without those lines and verify that it works fine?
Can we automate this check? 
* Generate phar
* using phar with some rule with some files

fain182:
I tried to generate the phar (with `make phar`) but it failed, even on the main branch..

But if I understand correctly the github action make a smoke test on the phar: 
https://github.com/phparkitect/arkitect/blob/main/.github/workflows/build.yml#L116

AlessandroMinoccheri:
The smoke test was done for that: test the phar generation.
@micheleorselli what do you think?

AlessandroMinoccheri:
Which error do you get @fain182 ? 
It's about `/tmp/..` permission, right?

bin-stub/phparkitect
```diff
@@ -7,28 +7,7 @@ use Arkitect\CLI\Check;
...
-        'You need to set up the project dependencies using Composer:' . PHP_EOL . PHP_EOL .
-        '    composer install' . PHP_EOL . PHP_EOL .
-        'You can learn all about Composer on https://getcomposer.org/.' . PHP_EOL
-    );
-
-    die(1);
-}
-
-require PHPARKITECT_COMPOSER_INSTALL;
+require __DIR__ . '/../vendor/autoload.php';
```
micheleorselli, line 10: AFAIK the only thing the above code is supposed to do is setting PHPARKITECT_COMPOSER_INSTALL as a way to say the package was installed via composer 🤔

To me it can be safely removed

On the other hand, I am not totally sure about adding this line: what is it effect?

fain182:
> Which error do you get @fain182 ? It's about `/tmp/..` permission, right?

the first error is:
```
bin/box.phar compile -c /tmp/arkitect/box.json
make: bin/box.phar: Permesso negato
make: *** [Makefile:26: phar] Errore 127
```
If I correct the permission, it says something strange about PHP versions...
Schermata del 2022-08-30 14-33-26

bin-stub/phparkitect
fain182: To me seems more like:
 - try to find and autoloader (why? it's always in the same place)
 - setup a warning if `composer install` has not been executed yet
 - include the autoload

The situation is different if we are searching the autoload of the project ouside of the phar, but I don't think it's this case.

micheleorselli: you are totally right... let's remove it then

micheleorselli:
> > Which error do you get @fain182 ? It's about `/tmp/..` permission, right?
> 
> the first error is:
> 
> ```
> bin/box.phar compile -c /tmp/arkitect/box.json
> make: bin/box.phar: Permesso negato
> make: *** [Makefile:26: phar] Errore 127
> ```
> 
> If I correct the permission, it says something strange about PHP versions... Schermata del 2022-08-30 14-33-26



> > Which error do you get @fain182 ? It's about `/tmp/..` permission, right?
> 
> the first error is:
> 
> ```
> bin/box.phar compile -c /tmp/arkitect/box.json
> make: bin/box.phar: Permesso negato
> make: *** [Makefile:26: phar] Errore 127
> ```
> 
> If I correct the permission, it says something strange about PHP versions... Schermata del 2022-08-30 14-33-26

If I remember correctly, the Box version we have works fine php 7.1, which is the version we use to generate the phar. Other 7.x version should work, 8.x version will not. More recent Box version will work with php 8, but they dropped the support for php 7.1

Long story short, if you want to generate the phar you need to have php 7

micheleorselli:
I tinkered a bit with it and for what I got we cannot just do a `require __DIR__ . '/../vendor/autoload.php';`

Context: we need to include the autoload file for phparkitect (not the one of the project we are analysing). As of now we do have two ways of using/installing phparkitect
- as a phar
- as a composer package 

When we use phparkitect as a phar the tool we use to package it takes care of embedding the autoload, so `require __DIR__ . '/../vendor/autoload.php';` will work  ✅ 

When we use phparkitect as a composer package the location of the phparkitect executable is configurable via composer json. It could be on the project root, project bin dir, vendor bin dir, and so on. In that case we need a way to guess where it is. 

From composer 2.2 (which I would say it covers the majority of cases) a variable is set by composer itself with the autoload path (`$_composer_autoload_path`). 

For previous composer versions a fallback mechanism is provided which tries to guess the bin location (not sure if it makes sense to support that)

wdyt @AlessandroMinoccheri @fain182?

fain182:
I think that's a good idea...
And I agree that supporting older composer version is excessive.

PR #289, opened 2022-08-29
added php 8.2 to ci tests

AlessandroMinoccheri:
Now it's broken, I think because the version of PHP is php8.2-dev

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
         "vimeo/psalm": "^4.6",
         "friendsofphp/php-cs-fixer": "^2.16",
         "phpunit/phpunit": "^7.5|^9.0",
-        "mikey179/vfsstream": "^1.6"
+        "mikey179/vfsstream": "^1.6",
+        "phpspec/prophecy": "^1.10"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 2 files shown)


fain182:
Maybe we need to change the minimum-stability? 
https://getcomposer.org/doc/04-schema.md#minimum-stability

AlessandroMinoccheri:
We can change it but I don't know if it's better to wait for the stable release at the moment. 
We have other PRs and issues with more value I think. Do you agree @fain182 ?

composer.json
```diff
@@ -38,7 +38,8 @@
         "vimeo/psalm": "^4.6",
         "friendsofphp/php-cs-fixer": "^2.16",
         "phpunit/phpunit": "^7.5|^9.0",
-        "mikey179/vfsstream": "^1.6"
+        "mikey179/vfsstream": "^1.6",
+        "phpspec/prophecy": "^1.10"
```
pfazzi, line 42: why?

pfazzi (approved):

composer.json
fain182: I suppose this error was the reason 
https://github.com/phparkitect/arkitect/runs/8292607729?check_suite_focus=true

PR #292, opened 2022-09-12
added filter option to run a specific rule

AlessandroMinoccheri:
In this PR I added the option to filter a specific rule.
In this step, the rule name must be the same as the filter.
I remove the spread operator to allow string keys for our rules.

Future steps:
* using a specific collection for our rules instead of an array
* allows filtering multiple rules like: `--filter=controller` can filter these: `controller_extends`,  `controller_rule_2` ecc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,7 +335,16 @@ phparkitect check --config=/project/yourConfigFile.php
 ```
 * `--target-php-version`: With this parameter, you can specify which PHP version should use the parser. This can be useful to debug problems and to understand if there are problems with a different PHP version.
 Supported PHP versions are: 7.1, 7.2, 7.3, 7.4, 8.0, 8.1
- * `--stop-on-failure`: With this option the process will end immediately after the first violation.
+* `--stop-on-failure`: With this option the process will end immediately after the first violation.
+* `--filter`: With this option you can run a specific rule with the same name example: `--filter=myRule`.
+To use this filter you can give a name to your rule like this:
+```php
+$rules['myRule'] = Rule::allClasses()
+    ->except('App\Controller\FolderController\*')
+    ->that(new ResideInOneOfTheseNamespaces('App\Controller'))
+    ->should(new HaveNameMatching('*Controller'))
+    ->because('we want uniform naming');
+```
 
 # Integrations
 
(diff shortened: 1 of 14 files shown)


src/CLI/Config.php
```diff
@@ -17,15 +16,37 @@ public function __construct()
...
+            return $this->filterRuleIntoClassSetRules($ruleFilter);
+        }
+
+        return $this->classSetRules;
+    }
+
+    private function filterRuleIntoClassSetRules(string $ruleFilter): array
+    {
+        /** @var ClassSetRules $classSetRules */
+        foreach ($this->classSetRules as $index => $classSetRules) {
```
micheleorselli, line 38: why don't we create a collection of classSetRules instead of using a plain array?

README.md
```diff
@@ -335,7 +335,16 @@ phparkitect check --config=/project/yourConfigFile.php
 ```
 * `--target-php-version`: With this parameter, you can specify which PHP version should use the parser. This can be useful to debug problems and to understand if there are problems with a different PHP version.
 Supported PHP versions are: 7.1, 7.2, 7.3, 7.4, 8.0, 8.1
- * `--stop-on-failure`: With this option the process will end immediately after the first violation.
+* `--stop-on-failure`: With this option the process will end immediately after the first violation.
+* `--filter`: With this option you can run a specific rule with the same name example: `--filter=myRule`.
+To use this filter you can give a name to your rule like this:
+```php
+$rules['myRule'] = Rule::allClasses()
```
micheleorselli, line 342: While this is the most straightforward implementation, I wonder if this can prevent further evolutions. Since we are touching the config file I would be careful in introducing stuff since deprecating it in the future could be problematic. What do you think?

src/CLI/Runner.php
```diff
@@ -21,11 +21,14 @@ class Runner
 
     /** @var ParsingErrors */
     private $parsingErrors;
+    /** @var string|null */
+    private $ruleFilter;
 
-    public function __construct(bool $stopOnFailure = false)
+    public function __construct(bool $stopOnFailure = false, ?string $ruleFilter = null)
```
micheleorselli, line 27: is the default value mandatory?

PR #293, opened 2022-09-13
Added runOnlyThis option to ruleBuilder

AlessandroMinoccheri:


src/CLI/Config.php
```diff
@@ -11,14 +11,32 @@ class Config
...
+        if ($this->runOnlyARule) {
+            return $this;
+        }
+
+        /** @var ArchRule $rule */
+        foreach ($rules as $rule) {
+            if ($rule->isRunOnlyThis()) {
+                $rules = [];
+                $rules[] = $rule;
+
```
fain182, line 34: maybe it's simpler to insert the return here and remove the runOnlyARule attribute?

tests/Unit/CLI/ConfigTest.php
```diff
@@ -31,4 +31,28 @@ public function test_it_should_create_config(): void
...
+            ->should(new HaveNameMatching('*Service'))
+            ->because('all services should be end name with Service')
+            ->runOnlyThis();
+
+        $config = new Config();
+        $config->add($classSet, ...[$rule1, $rule2]);
+
+        $this->assertInstanceOf(Config::class, $config);
+
+        $classSetRulesExpected[] = ClassSetRules::create($classSet, ...[$rule2]);
```
fain182, line 55: I suppose the spread operator of the one element array is a typo?

src/CLI/Config.php
AlessandroMinoccheri: You can call multiple times this method with different ClassSet, it's not the best approach I know but adding the `return` here I think it's not correct.
What do you think?

tests/Unit/CLI/ConfigTest.php
AlessandroMinoccheri: I think it's correct, I copied the previous test and it seems to work fine.

src/CLI/Config.php
fain182: oh, i get it now... you are right!

tests/Unit/CLI/ConfigTest.php
fain182: I was saying that:
`...[$rule2]` is the same as `$rule2`
but if you prefer to keep it like this for consistency with other tests it's ok, it's irrelevant
:slightly_smiling_face:

PR #295, opened 2022-10-03
Fix quotes in php-code (in README) to fix highlighting

jkrzefski:


README.md
```diff
@@ -122,7 +122,7 @@ $rules = Rule::allClasses()
 $rules = Rule::allClasses()
     ->that(new ResideInOneOfTheseNamespaces('App\Controller'))
     ->should(new DocBlockNotContains('@psalm-immutable'))
-    ->because('we don't want to enforce immutability');
+    ->because("we don't want to enforce immutability");
```
AlessandroMinoccheri, line 125: Good catch, but what do you think about using single quotes and escape characters into the string?

AlessandroMinoccheri (changes requested):

README.md
jkrzefski: I can do that as well, but since this code exists only for presentational purposes, I thought adding a backslash could be a bit distracting. Should I change it?

AlessandroMinoccheri: IMHO I would prefer the consistency with other lines using single quotes :)

jkrzefski: Done. I also found another line that used double quotes for the same reason. I replaced those, too.

AlessandroMinoccheri (approved):
LGTM! 
Thanks for your contribution! 👍

Issue #301, opened 2022-10-18
Check doc blocks at functions


Mysak007:
### Feature Request
I'm not sure if it is a bug.
I'm using  "phparkitect/phparkitect": "^0.2.32",

Please implement rule to check if docblock at function contains string.

#### Summary
I tried to implement rule 
```
// exception rule
	$rules[] = Rule::allClasses()
		->that(new ResideInOneOfTheseNamespaces('App'))
		->should(new DocBlockNotContains('@throws Exception'))
		->because('we don\'t want simple exceptions - use custom instead');
```

but it checks only docblock at classes, not at functions so when I have functions like this, I need to check it.

```
/**
	 * @throws Exception
	 * @throws ItemNotFoundException
	 */
	public function handleTitleCreate(TitleCreateParameters $parameters): Title
```

The same problem is when I want to check that entity doesn't use repository (IE: #[ORM\Entity(repositoryClass: ProductRepository::class)]

So when I use:
`// entity docblock doesn't contain repository
	$rules[] = Rule::allClasses()
		->that(new ResideInOneOfTheseNamespaces('App\Entity'))
		->should(new DocBlockContains('#[ORM\Entity(repositoryClass:'))
		->because('we don\'t want to use repository class in our entities');`
It doesn't work.

thanks a lot :)
Michal

AlessandroMinoccheri:
Hi @Mysak007, we only analyze the docblock on the class and not for every function.
So we can try to analyze the function's docblock.

AlessandroMinoccheri:
@Mysak007 we released the new version 0.3.0 
Pay attention, we changed rules for docBlock in: `ContainDocBlockLike` and `NotContainDocBlockLike`
Feel free to open new issue about this topic, at the moment I consider closed this issue :) 

Thanks!

PR #302, opened 2022-10-18
added parsing and checking functions docblocks

AlessandroMinoccheri:
This should fix this issue #301 
Is it a BC?

### DIFF
--- a/src/Analyzer/FileVisitor.php
+++ b/src/Analyzer/FileVisitor.php
@@ -47,12 +47,6 @@ class FileVisitor extends NodeVisitorAbstract
                 $this->classDescriptionBuilder->setAbstract(true);
             }
 
-            if (null !== $node->getDocComment()) {
-                /** @var Doc $docComment */
-                $docComment = $node->getDocComment();
-                $this->classDescriptionBuilder->setDocBlock($docComment->getText());
-            }
-
             foreach ($node->attrGroups as $attributeGroup) {
                 foreach ($attributeGroup->attrs as $attribute) {
                     $this->classDescriptionBuilder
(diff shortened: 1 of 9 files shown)


src/Analyzer/FileVisitor.php
```diff
@@ -133,6 +127,12 @@ public function enterNode(Node $node): void
             );
         }
 
+        if (null !== $this->classDescriptionBuilder && null !== $node->getDocComment()) {
```
micheleorselli, line 130: do we need the null check on `$this->classDescriptionBuilder`? I find it a bit confusing, as we are already using `$this->classDescriptionBuilder` a few lines above without any check

micheleorselli (commented):
Noice! Left a minor, non blocking comment

src/Analyzer/FileVisitor.php
AlessandroMinoccheri: I have got different errors without this check

micheleorselli: I am not sure I got what you mean but just out of curiosity: doesn't the check on line 28 prevent from having a null in `$this->classDescriptionBuilder` here?

And just to be super clear: all good from me, feel free to merge it

fain182 (approved):
:+1:

src/Analyzer/FileVisitor.php
AlessandroMinoccheri: The check on line 28 is inside an if that finish before the new check, this check is more generic

AlessandroMinoccheri: split the if @fain182 ? What is confusing in this part?

PR #304, opened 2022-10-27
renamed dockBlock expressions with a fluent english name

AlessandroMinoccheri:


fain182 (approved):
Lgtm 🔝

micheleorselli (commented):
In order to remain consistent with the other classes we should use the verb in 3rd plural form

all classes that `ContainDocBlockLike`
all classes that ... should `NotContainDocBlockLike`

PR #305, opened 2022-10-28
Handle attributes on enums (PHP8.1)

jerowork:
# Description
Just as with classes, PHPArkitect should handle attributes for enums as well, not resulting in a violation when attributes are present in the enum.

More information in the issue https://github.com/phparkitect/arkitect/issues/306

## Changes
- [x] Handle attributes for enums as well in `FileVisitor`


AlessandroMinoccheri (approved):
LGTM 👍

Issue #307, opened 2022-11-04
Several types of dependency are not detected


LuigiCardamone:
### Feature Request

Detect all types of dependency and add an option to detect all dependencies used on custom annotations.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

#### Summary
Hi All.

I created an Application level DTO that should depend only from Application and Domain using the rule `DependsOnlyOnTheseNamespaces` but I notice that several types of dependency from the framework are not detected.
I attach some examples.

Example 1 with custom annotation --> Not detected
```
use Symfony\Component\Validator\Constraints as Assert;

class ApplicationLevelDto
{
    /**
     * @Assert\NotBlank
     */
    public string|null $foo;
}
```

Example 2 with `@var` --> Not detected
```
use Symfony\Component\Validator\Constraints\NotBlank;

class ApplicationLevelDto
{
    /**
     * @var NotBlank
     */
    public $foo;
}
```

Example 3 with typed property --> Not detected
```
use Symfony\Component\Validator\Constraints\NotBlank;

class ApplicationLevelDto
{
    public NotBlank $foo;
}
```

It works only when I write this:
```
use Symfony\Component\Validator\Constraints\NotBlank;

class ApplicationLevelDto
{
    public NotBlank $foo;

    public function __construct()
    {
        $this->foo = new NotBlank(); //Violation detected in this line
    }
}
```

Example 1 may be ok since that annotation is just adding extra information and the class may work even without that dependency: maybe could it be an optional check?
Example 2 could be useful for legacy project.
Example 3 seems a bug: what you think?

LuigiCardamone:
Hi @AlessandroMinoccheri, I attached my config in the description of the issue

AlessandroMinoccheri:
@LuigiCardamone we released version 0.3.1 with the fix for your third problem about the typed property.
Asap we will fix the others!

mmenozzi:
Hi @AlessandroMinoccheri, with a configuration like the following:

```
    $config->add(
        ClassSet::fromDir(__DIR__ . '/src'),
        Rule::allClasses()
            ->that(new ResideInOneOfTheseNamespaces('MyProject\Domain'))
            ->should(
                new NotHaveDependencyOutsideNamespace('MyProject\Domain')
            )
            ->because('domain should not have external dependencies'),
    );
```

Version `0.3.1` of pkparkitect reports a property declaration like this:

```
namespace MyProject\Domain;

[...]

PR #309, opened 2022-11-12
parsing property type into docblocks

AlessandroMinoccheri:
This PR can fix the second docblock problems explained in this issue: #307 about @var --> Not detected

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,8 @@
         "symfony/polyfill-php80": "^1.20",
         "nikic/php-parser": "~4",
         "webmozart/assert": "^1.9",
-        "ext-json": "*"
+        "ext-json": "*",
+        "phpstan/phpdoc-parser": "^1.2"
     },
     "require-dev": {
         "roave/security-advisories": "dev-master",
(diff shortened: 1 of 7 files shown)


src/Analyzer/DocBlockNameResolver.php
```diff
@@ -0,0 +1,71 @@
...
+            $docComment = $node->getDocComment();
+
+            $tokens = $lexer->tokenize($docComment->getText());
+            $tokenIterator = new TokenIterator($tokens);
+            $phpDocNode = $phpDocParser->parse($tokenIterator);
+
+            foreach ($phpDocNode->getVarTagValues() as $tagValue) {
+                $type = $this->resolveName(new Node\Name((string) $tagValue->type), Use_::TYPE_NORMAL);
+                $node->type = $type;
+                break;
```
AlessandroMinoccheri, line 46: This break I don't know if this is correct or not, it seems that the first element is the type that we want.  I haven't found cases with more elements, but to avoid problems  I preferred to exit immediately.

micheleorselli (approved):

PR #310, opened 2022-11-14
Fix bug with typed properties and scalar types (#308)

LuigiCardamone:
This fix the bug reported by @mmenozi on issue (#308) about scalar types in typed properties. With this fix scalar types will be not detected as dependencies.


AlessandroMinoccheri (approved):
LGTM 👍 
Thanks @LuigiCardamone

Issue #311, opened 2022-11-21
Array shape docblocks causing issues since 0.3.3


ostrolucky:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        |no
| Library Version | 0.3.3
| PHP version     | 8.1.11

#### Summary

After updating phparkitect, output is ridden with these kind of errors
```
Parse Error: Haushalt\Bl\DomainBundle\Application\ItemData\Helper\array<int, Product> is not a valid namespace definition#0 /vagrant/fe-bl-domain/vendor/phparkitect/phparkitect/src/Analyzer/ClassDependency.php(17): Arkitect\Analyzer\FullyQualifiedClassName::fromString()
#1 /vagrant/fe-bl-domain/vendor/phparkitect/phparkitect/src/Analyzer/FileVisitor.php(156): Arkitect\Analyzer\ClassDependency->__construct()
#2 /vagrant/fe-bl-domain/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(200): Arkitect\Analyzer\FileVisitor->enterNode()
#3 /vagrant/fe-bl-domain/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(114): PhpParser\NodeTraverser->traverseArray()
#4 /vagrant/fe-bl-domain/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(223): PhpParser\NodeTraverser->traverseNode()
#5 /vagrant/fe-bl-domain/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(114): PhpParser\NodeTraverser->traverseArray()
#6 /vagrant/fe-bl-domain/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(223): PhpParser\NodeTraverser->traverseNode()
#7 /vagrant/fe-bl-domain/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(91): PhpParser\NodeTraverser->traverseArray()
#8 /vagrant/fe-bl-domain/vendor/phparkitect/phparkitect/src/Analyzer/FileParser.php(74): PhpParser\NodeTraverser->traverse()
#9 /vagrant/fe-bl-domain/vendor/phparkitect/phparkitect/src/CLI/Runner.php(57): Arkitect\Analyzer\FileParser->parse()
[...]

PR #312, opened 2022-11-21
added a try catch to avoid errors on parsing docblocks

AlessandroMinoccheri:
This PR solve #311


micheleorselli (approved):
If we want to fix this bug asap it is fine to wrap the `addDependency` method like this, but it should be temporary

PR #314, opened 2022-11-22
Add baseline feature

pfazzi:
If there are a lot of violations in your codebase and you can't at the moment fix them, 
you can use the baseline feature to instruct the tool to ignore past violations.

To create a baseline file, run the `check` command with the `set-baseline` parameter as follows:

```
phparkitect check --set-baseline=my-baseline.json
```

It will produce a json file with the current list of violations.  

To instruct the tool to use the baseline file, run the `check` command with the `use-baseline` parameter as follows:

```
phparkitect check --use-baseline=my-baseline.json
```

## TODO

- [x] change `--set-baseline` in `--generate-baseline`
- [x] create baseline with a standard filename if is not given
- [x] use baseline file if is present
- [x] prints if a baseline is used
- [x] adds `--skip-baseline` option
- [x] fix documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,21 +60,33 @@ By default, a progress bar will show the status of the ongoing analysis.
 
 ### Using a baseline file
 
-If there are a lot of violations in your codebase and you can't at the moment fix them, 
+If there are a lot of violations in your codebase and you can't fix them now, 
 you can use the baseline feature to instruct the tool to ignore past violations.
 
-To create a baseline file, run the `check` command with the `set-baseline` parameter as follows:
+To create a baseline file, run the `check` command with the `generate-baseline` parameter as follows:
 
+```
+phparkitect check --generate-baseline
+```
+This will create a `phparkitect-baseline.json`, if you want a different file name you can do it with:
 ```
 phparkitect check --generate-baseline=my-baseline.json
 ```
 
 It will produce a json file with the current list of violations.  
 
-To instruct the tool to use the baseline file, run the `check` command with the `use-baseline` parameter as follows:
+If is present a baseline file with the default name will be used automatically.
+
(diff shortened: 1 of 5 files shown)


AlessandroMinoccheri (approved):

pfazzi:
> I was thinking that maybe could be a little more usable if:
> 
> * we decide a standard baseline filename
> * if the file is present is used even without any option specified
> * the `--use-baseline` is used only when you want to use a non-standard filename (is needed? maybe we can avoid it)
> * we add a `--skip-baseline` option to avoid to use the default baseline
> 
> The idea is that if a baseline is present is much more comfortable not to need to say it explicitely every time..

Make sense, we could do it in a next step. What do you think?

micheleorselli:
@fain182 @pfazzi for the sake of consistency it could be worth checking what other tools are doing, what do you think?

pfazzi:
> @fain182 @pfazzi for the sake of consistency it could be worth checking what other tools are doing, what do you think?

I basically copied the same behavior of psalm :D

fain182:
I am not sure about improving it in the future because it would be a breaking change..

Phpstan uses `--generate-baseline` that is slightly more explicit, and if you don't pass any filename it uses the default filename.

README.md
```diff
@@ -58,6 +58,25 @@ phparkitect check --config=/project/yourConfigFile.php
 
 By default, a progress bar will show the status of the ongoing analysis.
 
+### Using a baseline file
```
pfazzi, line 61: @fain182 reisci a dare na botta al readme?

fain182: sissì, l'ho messo tra i TODO nella descrizione della PR :slightly_smiling_face:

PR #316, opened 2022-11-22
Implement no class feature

pfazzi:


src/Rules/RuleBuilder.php
```diff
@@ -69,9 +79,9 @@ public function classesToBeExcluded(string ...$classesToBeExcluded): self
         return $this;
     }
 
-    public function setRunOnlyThis(): self
+    public function negateShoulds(): self
```
AlessandroMinoccheri, line 82: Why did you do this change?

AlessandroMinoccheri (commented):
I would suggest to add a description of this new behavior into the README file

src/Expression/NegateDecorator.php
```diff
@@ -0,0 +1,57 @@
...
+declare(strict_types=1);
+
+namespace Arkitect\Expression;
+
+use Arkitect\Analyzer\ClassDescription;
+use Arkitect\Rules\Violation;
+use Arkitect\Rules\ViolationMessage;
+use Arkitect\Rules\Violations;
+
+class NegateDecorator implements Expression
```
micheleorselli, line 11: What about naming it just `Not`? I could also simplify other expressions

src/Rules/Rule.php
```diff
@@ -9,4 +9,9 @@ public static function allClasses(): AllClasses
     {
         return new AllClasses();
     }
+
+    public static function noClass(): NoClass
```
micheleorselli, line 13: what about naming it NoClasses, for consistency with AllClasses?

src/Rules/RuleBuilder.php
```diff
@@ -39,6 +45,10 @@ public function addThat(Expression $that): self
 
     public function addShould(Expression $should): self
     {
+        if ($this->negateShoulds) {
+            $should = new NegateDecorator($should);
+        }
```
micheleorselli, line 50: Maybe I am too low on ☕  but I can't wrap my head around this, would you mind explaining it to me? 
If the should method in NoClass is defined as follow

```
$this->ruleBuilder->addShould(new NegateDecorator($expression));
```

why do we need to add another `new NegateDecorator` here?

src/Expression/NegateDecorator.php
pfazzi: Totally agree!

src/Rules/Rule.php
pfazzi: Because it would not be correct English. It's the same in Italian "tutte le classi"/"nessuna classe".

micheleorselli: I am far from being an expert but I think also NoClasses is correct. Looking at [java's ArchUnit](https://www.archunit.org/userguide/html/000_Index.html) they decided to used it, eg. 

```java
ArchRule rule = ArchRuleDefinition.noClasses()
    .that().resideInAPackage("..service..")
    .should().accessClassesThat().resideInAPackage("..controller..");

rule.check(importedClasses);
```

PR #317, opened 2022-11-27
Parse custom tags

AlessandroMinoccheri:
In this PR the code can parse custom tags in property docblocks.


micheleorselli (approved):
Looks good to me!

Issue #329, opened 2023-01-01
Bug in `except` method with regex input


smortexa:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | yes
| Library Version | 0.3.9
| PHP version     | 8.1.8

#### Summary

After upgrading the library  to version `0.3.9` the following code doesn't work:
`->except('App\Providers\(Auth|Event|Route|Horizon)ServiceProvider')`
but this works:
`->except('App\Providers\AuthServiceProvider', 'App\Providers\EventServiceProvider')`

#### Current behavior

```php
App\Providers\AuthServiceProvider has 1 violations
  should extend Illuminate\Support\ServiceProvider because we use Laravel framework!

App\Providers\EventServiceProvider has 1 violations
  should extend Illuminate\Support\ServiceProvider because we use Laravel framework!

App\Providers\HorizonServiceProvider has 1 violations
  should extend Illuminate\Support\ServiceProvider because we use Laravel framework!

App\Providers\RouteServiceProvider has 1 violations
  should extend Illuminate\Support\ServiceProvider because we use Laravel framework!
```

#### How to reproduce

Use `providers` rule in https://github.com/smortexa/laravel-arkitect:
```php
Rule::allClasses()
            ->except('App\Providers\(Auth|Event|Route|Horizon)ServiceProvider')
            ->that(new ResideInOneOfTheseNamespaces('App\Providers'))
            ->should(new Extend('Illuminate\Support\ServiceProvider'))
            ->because('we use Laravel framework!');
```

#### Expected behavior

No violation report.

AlessandroMinoccheri:
Hi @smortexa, in previous releases works fine your example?

smortexa:
@AlessandroMinoccheri Yes. Until this MR: https://github.com/smortexa/laravel-arkitect/pull/15

AlessandroMinoccheri:
@fain182 could it be related to the last work on regexp?

fain182:
Yes, it could be.. I work on it

fain182:
@smortexa actually the regex support was added by error... we support only shell wildcard ( `*` and` ?`).

With the new release we added an explicit error message for regex.

PR #327, opened 2022-12-17
Avoid partial name matching

AlessandroMinoccheri:

PR #330, opened 2023-01-01
Restore support in regex for namespaces and path

fain182:
We thought to permits only unix shell wildcard for defining namespaces.

In the PR #188 we accidentally added the support for some regexes.

With the PR #327 we made the check a little bit more strict, to avoid bug with partial name matching, but broke the support for some regexes #329 .

We have 3 options:

 1 - support both regex and unix wildcard in the same method (this PR fix the problem that we saw now, but I fear there will be a lot more problems in the future)
 2 - support only unix wildcard and remove the support for regexes
 3 - remove the support for unix wildcard and make a major version that uses only regexes

@micheleorselli @AlessandroMinoccheri what do you think?

I am thinking about solution 2...

### DIFF
--- a/src/Analyzer/ClassDescription.php
+++ b/src/Analyzer/ClassDescription.php
@@ -14,9 +14,6 @@ class ClassDescription
     /** @var list<FullyQualifiedClassName> */
     private $interfaces;
 
-    /** @var string */
-    private $fullPath;
-
     /** @var ?FullyQualifiedClassName */
     private $extends;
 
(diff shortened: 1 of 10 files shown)


AlessandroMinoccheri:
I have different ideas about the problem but at the moment I think that solution 1 could be the smallest step to solving a problem.
We don't really know if devs use regex or wildcard or both.

We can fix the actual problem and after that, we can try to use a value object instead of a string to keep things testable and to support different strategies.

What do you think @micheleorselli?

micheleorselli:
The initial idea was to support wildcards, the fact that regexes slipped into the codebase should have not happened in my opinion. My proposal is to stick with wildcards for version 1.0 and then revise this decision after that, what do you think?

AlessandroMinoccheri:
Ok, @micheleorselli supporting only wildcards means that if someone is using regexes we will break its code, right?

micheleorselli:
> Ok, @micheleorselli supporting only wildcards means that if someone is using regexes we will break its code, right?

Potentially yes, but we are talking about a feature shipped by mistake, so I think it would not have a big impact

AlessandroMinoccheri:
I love the idea of the error message @fain182!
So I think we can proceed to remove the Regex support accepting only wildcards!

PR #333, opened 2023-01-12
Fix remove violations from baseline

hectorespert:
On a PHP 7.4 project, after define a baseline, if a violation occurs, all violations are showed, the new violations and the violations defined in the baseline instead to show only new violations.

### DIFF
--- a/src/Rules/Violations.php
+++ b/src/Rules/Violations.php
@@ -108,13 +108,13 @@ class Violations implements \IteratorAggregate, \Countable, \JsonSerializable
 
     public function remove(self $violations): void
     {
-        $this->violations = array_udiff(
+        $this->violations = array_values(array_udiff(
             $this->violations,
             $violations->violations,
-            function (Violation $a, Violation $b): int {
-                return $a == $b ? 0 : 1;
+            static function (Violation $a, Violation $b): int {
+                return $a->isEqualsTo($b) ? 0 : -1;
             }
-        );
+        ));
     }
 
     public function jsonSerialize(): array
--- a/src/Rules/Violations.php
+++ b/src/Rules/Violations.php
@@ -112,7 +112,7 @@ class Violations implements \IteratorAggregate, \Countable, \JsonSerializable
             $this->violations,
             $violations->violations,
             static function (Violation $a, Violation $b): int {
-                return $a->isEqualsTo($b) ? 0 : -1;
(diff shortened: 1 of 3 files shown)


AlessandroMinoccheri:
Hi @hectorespert, 
thanks for this pull request!
can you please fix the coding standards?

hectorespert:
Fixex

src/Rules/Violations.php
```diff
@@ -108,13 +108,13 @@ public function toArray(): array
 
     public function remove(self $violations): void
     {
-        $this->violations = array_udiff(
+        $this->violations = array_values(array_udiff(
```
AlessandroMinoccheri, line 111: is it possible to simplify this code for you?

hectorespert: I could remove the `array_values` method. It is only added because `array_udiff` maintain the old array keys for the violations.

AlessandroMinoccheri: I mean to refactor the function to be more explicit. Now we have your test and we can improve the code :) 
We can do it also in another PR.

hectorespert: In that case, I prefer to not modify that in this PR, if a new version with this fix could be released.

AlessandroMinoccheri (approved):
LGTM 👍

src/Rules/Violations.php
```diff
@@ -108,13 +108,13 @@ public function toArray(): array
...
             $this->violations,
             $violations->violations,
-            function (Violation $a, Violation $b): int {
-                return $a == $b ? 0 : 1;
+            static function (Violation $a, Violation $b): int {
+                return $a->isEqualsTo($b) ? 0 : -1;
```
micheleorselli, line 115:
```suggestion
                return $a <=> $b;
```

this should do the trick without a custom comparison function. would you mind trying it?

micheleorselli (changes requested):
Thank you for working on this! 🙇 I would ask to do one last experiment before merging, using the spaceship operator <=> instead of a custom comparison function

src/Rules/Violations.php
hectorespert: I changed that to use the spaceship operator

micheleorselli (approved):
Thank you @hectorespert 🙇

Issue #336, opened 2023-01-13
Add a configuration to skip PHPDoc Custom Tag detection


LuigiCardamone:
### Feature Request

Add a configuration to skip PHPDoc Custom Tag detection since in some projects annotation like `@JMS\Expose` cannot be easily removed. I think that skipping this check is not wrong since these annotations are not real dependencies. 

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

I will try to make a PR for this feature.

AlessandroMinoccheri:
Closed by #337

PR #337, opened 2023-01-13
Add a configuration to skip DocBlock Custom Tag detection (#336)

LuigiCardamone:
This PR implements the feature described in this issue: https://github.com/phparkitect/arkitect/issues/336

To skip the check you need to add this configuration: 

```
return static function (Config $config): void {
    $config->setParseDocBlockCustomTags(false);
    //...
}
```

No breaking change is added: if the configuration is not modified the default behavior is used (ParseDocBlockCustomTags=true).

### DIFF
--- a/src/Analyzer/NameResolver.php
+++ b/src/Analyzer/NameResolver.php
@@ -31,6 +31,9 @@ class NameResolver extends NodeVisitorAbstract
     /** @var bool Whether to replace resolved nodes in place, or to add resolvedNode attributes */
     protected $replaceNodes;
 
+    /** @var bool Whether to parse DocBlock Custom Annotations */
+    protected $parseCustomAnnotations;
+
     /**
      * Constructs a name resolution visitor.
      *
@@ -40,6 +43,7 @@ class NameResolver extends NodeVisitorAbstract
      *  * replaceNodes (default true): Resolved names are replaced in-place. Otherwise, a
      *    resolvedName attribute is added. (Names that cannot be statically resolved receive a
      *    namespacedName attribute, as usual.)
+     *  * parseCustomAnnotations (default true): Whether to parse DocBlock Custom Annotations.
      *
      * @param ErrorHandler|null $errorHandler Error handler
      * @param array             $options      Options
@@ -49,6 +53,7 @@ class NameResolver extends NodeVisitorAbstract
         $this->nameContext = new NameContext($errorHandler ?? new ErrorHandler\Throwing());
         $this->preserveOriginalNames = $options['preserveOriginalNames'] ?? false;
         $this->replaceNodes = $options['replaceNodes'] ?? true;
+        $this->parseCustomAnnotations = $options['parseCustomAnnotations'] ?? true;
     }
 
     /**
@@ -148,7 +153,7 @@ class NameResolver extends NodeVisitorAbstract
                 break;
             }
 
-            if (!($node->type instanceof FullyQualified)) {
+            if ($this->parseCustomAnnotations && !($node->type instanceof FullyQualified)) {
                 foreach ($phpDocNode->getTags() as $tagValue) {
                     if ('@' === $tagValue->name[0] && false === strpos($tagValue->name, '@var')) {
                         $customTag = str_replace('@', '', $tagValue->name);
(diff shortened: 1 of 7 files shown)


src/Analyzer/NameResolver.php
```diff
@@ -148,14 +153,16 @@ public function enterNode(Node $node)
                 break;
             }
 
-            if (!($node->type instanceof FullyQualified)) {
-                foreach ($phpDocNode->getTags() as $tagValue) {
-                    if ('@' === $tagValue->name[0] && false === strpos($tagValue->name, '@var')) {
-                        $customTag = str_replace('@', '', $tagValue->name);
-                        $type = $this->resolveName(new Node\Name($customTag), Use_::TYPE_NORMAL);
-                        $node->type = $type;
+            if ($this->parseDocBlockCustomTags) {
```
AlessandroMinoccheri, line 156: Is it possible to avoid another nested level?

AlessandroMinoccheri (changes requested):
Can you please change the readme file documenting how to skip DocBlock Custom Tag detection ?

src/Analyzer/NameResolver.php
LuigiCardamone: done! I also updated the docs.

tests/Unit/Analyzer/FileVisitorTest.php
```diff
@@ -730,4 +730,34 @@ public function getBookList(): QueryBuilder;
...
+
+        /** @var FileParser $fp */
+        $fp = FileParserFactory::createFileParser(TargetPhpVersion::create('8.1'), false);
+        $fp->parse($code, 'relativePathName');
+
+        $cd = $fp->getClassDescriptions();
+
+        $violations = new Violations();
+
+        $notHaveDependencyOutsideNamespace = new DependsOnlyOnTheseNamespaces('MyProject\AppBundle\Application');
```
AlessandroMinoccheri, line 758: Can you please change the var name with the correct one?

LuigiCardamone: @AlessandroMinoccheri done!

AlessandroMinoccheri (approved):
LGTM 👍

LuigiCardamone:
> I am not sure if we should skip entirely the parsing of dependencies in a dobclock, since they are "soft" dependencies. What is your opinion on that?
> 
> If we stick to exclude only custom docblock may I suggest to change the name of the config method to make it more explicit? Something like `$config->skipParsingAnnotations` or `$config->doNotParseAnnotations()`

In my opinion with DocBlocks you can handle two complete different cases:

1. Express the type of the variable:
```php
    /**
     * @var QueryBuilder
     */
    private $foo;
```
that is just the legacy version of `private QueryBuilder $foo`.

2. Give additional information about that variable:
```php
    /**
     * @JMS\Expose
     */
    private string $foo;
```
The class may work event without that information (that is used outside).

I guess that violations about Case 1 are much more critical than Case 2. In our project we cannot easily remove dependency about Case 2. We need an option to skip violations of Case 2 (like `@JMS\Expose` and similar).

How would you call such an option?
I used the term "customTag" since a `$customTag` variable was already used in `NameResolver.php`.
I am afraid that `skipParsingAnnotations` can be misleading with Case 1. What about `skipParsingCustomAnnotations` or `parseOnlyTypingAnnotations`?

micheleorselli:
ok, let's keep the current PR behaviour. regarding the name `skipParsingCustomAnnotations` sounds fine to me

LuigiCardamone:
> ok, let's keep the current PR behaviour. regarding the name `skipParsingCustomAnnotations` sounds fine to me

@micheleorselli I just updated the code, the test and the documentation.

The GitHub check that failed is for something not related to this issue: "Composer 2.3.0 dropped support for PHP <7.2.5 and you are running 7.1.33-51+ubuntu20.04.1+deb.sury.org+1, please upgrade PHP or use Composer 2.2 LTS via "composer self-update --2.2". Aborting."

AlessandroMinoccheri:
@LuigiCardamone it's Github/composer issue, I run actions, and now it's totally green.

Issue #339, opened 2023-01-14
PHPArkitect reports array as class


javaDeveloperKid:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | yes/no
| Library Version | 0.3.10 
| PHP version     | 8.1.0

Given this config:
```php
        $rules[] = Rule::allClasses()
            ->that(new ResideInOneOfTheseNamespaces('App\Domain'))
            ->should(new NotHaveDependencyOutsideNamespace('App\Domain')
            ->because('we want protect our domain');
```
the following class
```php
Class MyClass
{
    private array $field1;

    public function __construct(array $field1)
    {
        $this->field1 = $field1;
    }
}
```
is reporting
```sh
App\Domain\MyClass has 1 violations
  depends on array, but should not depend on classes outside namespace App\Domain because we want protect our domain (on line 3)
```

AlessandroMinoccheri:
Closed by #340

PR #340, opened 2023-01-15
considering array type as a scalar type

AlessandroMinoccheri:
This PR solves this issue: #339

### DIFF
--- a/src/Analyzer/FileVisitor.php
+++ b/src/Analyzer/FileVisitor.php
@@ -140,7 +140,7 @@ class FileVisitor extends NodeVisitorAbstract
                 return;
             }
 
-            if ($this->isScalarType($node->type->toString())) {
+            if ($this->isBuiltInType($node->type->toString())) {
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


fain182 (approved):

src/Analyzer/FileVisitor.php
```diff
@@ -245,6 +245,6 @@ private function addParamDependency(Node\Param $node): void
 
     private function isScalarType(string $typeName): bool
     {
-        return \in_array($typeName, ['bool', 'int', 'float', 'string']);
+        return \in_array($typeName, ['bool', 'int', 'float', 'string', 'array']);
```
micheleorselli, line 248: I think we should expand this check to take into account all php's internal types (`resource`, `callable`, but also `iterable` or `Generator` and stuff like that) otherwise we are gonna face this problem again (but we should try to write a test to prove that). 

We could use [gettype](https://www.php.net/manual/en/function.gettype.php), minding that `object` is a type that can carry dependencies. Also, the method name `isScalarType` is misleading, I would rename it to `isInternalType`

AlessandroMinoccheri: I love the idea of renaming the function to `isInternalType `
However, I tried to add all the internal types manually.
Using `gettype` can be another good solution with some differences, for example:
`gettype` returns `integer ` instead of `int` and we need to match these types of values also.

LuigiCardamone: In my original implementation I focused only on scalar types since arrays requires a special treatment: your should distinguish between typed arrays (that may have dependencies) and untyped arrays (that never have dependecies).
Maybe it can be useful to add a test with both kind of arrays.

P.S.
Another name for `isInternalType` can be `isBuiltInType` since the term "built-in" is used in the PHP documentation https://www.php.net/manual/en/language.types.intro.php

AlessandroMinoccheri: Yeah `isBuiltInType` what do you think @micheleorselli, is it good enough at the moment?

micheleorselli (approved):

PR #352, opened 2023-02-09
Add error message for regex

fain182:
I added the error message as an exception.
The message is not perfectly aligned, but I don't know how to improve this cleanly.

Schermata del 2023-02-09 15-39-51

I took the opportunity to do some other little refactorings:

- the ClassDescriptionBuilder now is more a builder
- the Extends Rule now supports wildcard in the argument
- I simplified the Rule tests a lot.. I think we could use the same pattern in the other unit tests of rules

### DIFF
--- a/src/Analyzer/FullyQualifiedClassName.php
+++ b/src/Analyzer/FullyQualifiedClassName.php
@@ -3,6 +3,8 @@ declare(strict_types=1);
 
 namespace Arkitect\Analyzer;
 
+use Arkitect\Exceptions\InvalidPatternException;
+
 class FullyQualifiedClassName
 {
     /** @var PatternString */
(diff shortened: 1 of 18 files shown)


AlessandroMinoccheri (approved):
LGTM 👍

micheleorselli (approved):

src/Analyzer/FullyQualifiedClassName.php
```diff
@@ -28,11 +30,19 @@ public function toString(): string
...
+            throw new InvalidPatternException("'$pattern' is not a valid class or namespace pattern. Regex are not allowed, only * and ? wildcard.");
+        }
+
         return $this->class->matches($pattern);
     }
 
     public function matches(string $pattern): bool
     {
+        if ($this->isNotAValidPattern($pattern)) {
+            throw new InvalidPatternException("'$pattern' is not a valid class or namespace pattern. Regex are not allowed, only * and ? wildcard.");
```
dbu, line 43: i just tried something with latest main and this breaks what worked before for us.

we add rules like this:

    addRules($config, 'Client', ['App\\Client\\SomeClient\\*', ...SPECIAL_RULES_NAMESPACES]);

it now complains that this is a regex.


    analyze class set /usr/src/app/src/Client

        0/1315 [>---------------------------]   0%'App\Client\SomeClient\*' is not a valid class or namespace pattern. Regex are not allowed, only * and ? wildcard.

when i comment out this line, everything still works. is the pattern checker incorrect? or how should i separate namespaces?

Issue #344, opened 2023-01-23
Violations on typed arrays are not detected


LuigiCardamone:
### Bug Report

Violation on typed array are not detected. If I have an array of object with Type Foo, violations related to Foo are not detected.

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Library Version | 0.3.12

#### Summary

I have a failing test:

```php
public function test_it_handles_typed_arrays(): void
    {
        $code = <<< 'EOF'
<?php
namespace Domain\Foo;

use Domain\Bar;
use Application\MyDto;

class MyClass
{
    /**
     * @var array<int, MyDto>
     */
    private array $dtoList;

    /**
     * @param array<int, MyDto> $dtoList
     */
    public function __construct(Bar $bar, array $dtoList)
    {
        $this->dtoList = $dtoList;
    }

    /**
     * @return array<int, MyDto>
     */
    public function getList(): array
    {
        return $this->dtoList;
    }
}
EOF;

        /** @var FileParser $fp */
        $fp = FileParserFactory::createFileParser(TargetPhpVersion::create('7.1'));
        $fp->parse($code, 'relativePathName');
        $cd = $fp->getClassDescriptions();

        $violations = new Violations();

        $dependsOnTheseNamespaces = new DependsOnlyOnTheseNamespaces('Domain');
        $dependsOnTheseNamespaces->evaluate($cd[0], $violations, 'we want to add this rule for our software');

        $this->assertCount(3, $violations);
    }
```

Typed array may be defined using different format:
- `@var array<int, MyDto>`
- `@var MyDto[]`
- `@var list<MyDto>`

LuigiCardamone:
@AlessandroMinoccheri do you have any idea on how we can do this? Since array definitions may be nested could make sense to search for some external library?

AlessandroMinoccheri:
@LuigiCardamone I have an idea but would like to explore more possible solutions.
My idea is to create a RegEx inside `NameResolver.php` where we parse docComments to get array inner values.
At the moment our code creates this dependency:
`Domain\Foo\array<int, MyDto>`

So we must understand that the string contains an `array` type with types.

At the moment I have no other solutions but I am opened to other ways.

cc @micheleorselli

LuigiCardamone:
If we want to cover all the cases, maybe we can give a look at this library: https://github.com/phpstan/phpdoc-parser .
[...]

PR #355, opened 2023-02-13
Handle typed arrays

LuigiCardamone:
This PR adds support for types array (see #344 ) on class properties. Support for method params and return types will be done next.


src/Analyzer/NameResolver.php
```diff
@@ -341,4 +358,24 @@ private function resolveType($node)
...
         return $node;
     }
+
+    private function getPhpDocNode(Stmt\Property $node): ?PhpDocNode
+    {
+        if (null === $node->getDocComment()) {
+            return null;
+        }
+
+        $lexer = new Lexer();
```
LuigiCardamone, line 368: @AlessandroMinoccheri to avoid creating this objects every time we can add `Lexer` `PhpDocParser` and `TokenIterator` as class attributes: what you think?

AlessandroMinoccheri: Yes, we can try to do it.

src/Analyzer/NameResolver.php
```diff
@@ -341,4 +388,52 @@ private function resolveType($node)
...
+
+    private function getArrayItemType(TypeNode $typeNode): ?string
+    {
+        $arrayItemType = null;
+
+        if ($typeNode instanceof GenericTypeNode) {
+            if (1 === \count($typeNode->genericTypes)) {
+                $arrayItemType = (string) $typeNode->genericTypes[0];
+            } elseif (2 === \count($typeNode->genericTypes)) {
+                $arrayItemType = (string) $typeNode->genericTypes[1];
```
AlessandroMinoccheri, line 423: Why do you take the second parameter if the count is 2?

LuigiCardamone: When count is 2 we are handling the case `array<int, MyDto>` in which `$typeNode->genericTypes[0]` is the key of the array (in this case `int`) and `$typeNode->genericTypes[1]` is the the type of the item (in this case `MyDto`). This code makes sense since we already checked that the current node is an `array`.

LuigiCardamone: Some extra information.
When count is 1, we are handling case `list<MyDto>` (and also `array<MyDto>` but this is not a 100% correct syntax).
When `$typeNode instanceof ArrayTypeNode` we are handling the case `MyDto[]` which is called "legacy syntax" since with PHPStan and PSalm is not longer used

AlessandroMinoccheri: Ok, now is more clear, I don't know if there is a way to explicit it in the code without a lot of comments.

LuigiCardamone: The 3 cases are explicit in the test cases. If you prefer I can add a short comment for each case.

AlessandroMinoccheri: For me a small comment can work fine, I was thinking to change something in the code to be more explicit like a function with a specific name instead of directly the `if` condition

LuigiCardamone: Using a function will be something like that:
```
if ($this->isTypedArrayNodeWithListSyntax($typeNode)) {
    $arrayItemType = (string) $typeNode->genericTypes[0];
}
```

but the knowledge about `$typeNode->genericTypes` will be splitted in many points increasing code readability. In this case I think that a small comment is better.

tests/Unit/Analyzer/FileVisitorTest.php
```diff
@@ -825,4 +825,289 @@ public function __construct(array $field1)
...
+namespace Domain\Foo;
+
+use Application\MyDto;
+
+class MyClass
+{
+    /**
+     * @param list<MyDto> $dtoList
+     */
+    public function __construct(array $dtoList)
```
micheleorselli, line 964: not sure it is worth it, but what if we add a test case with multiple param?

micheleorselli (approved):
Thank you @LuigiCardamone!

src/Analyzer/NameResolver.php
AlessandroMinoccheri: A small comment is better I think, good enough at the moment @LuigiCardamone

tests/Unit/Analyzer/FileVisitorTest.php
LuigiCardamone: @micheleorselli yes, it can helps. I just added a test with a method with multiple params: 2 primitive types (one with native type and one with PHPDoc type) and 2 array types (one with a violation and one with a class allowed in the rules).

AlessandroMinoccheri (approved):
LGTM 👍 
Outstanding @LuigiCardamone!

PR #360, opened 2023-02-17
Fixed "IsFinal" for abstract classes and interfaces

sebastianstucke87:
Fixed `\Arkitect\Expression\ForClasses\IsFinal` because:
- abstract classes can not be final
- interfaces can not be final


AlessandroMinoccheri (approved):
LGTM!

PR #364, opened 2023-03-06
sort violations to avoid unnecessary diffs in the baseline

dbu:
in our project, we noticed that the order of entries when generating the baseline varies randomly. 

this changes it to sort all entries in the baseline, making it possible to see what the actual changes in the baseline are.

### DIFF
--- a/src/Rules/Violation.php
+++ b/src/Rules/Violation.php
@@ -9,11 +9,11 @@ class Violation implements \JsonSerializable
     /** @var string */
     private $fqcn;
 
-    /** @var string */
-    private $error;
-
     /** @var int|null */
     private $line;
+
+    /** @var string */
+    private $error;
 
     public function __construct(string $fqcn, string $error, ?int $line = null)
     {
--- a/src/Rules/Violations.php
+++ b/src/Rules/Violations.php
@@ -117,6 +117,13 @@ class Violations implements \IteratorAggregate, \Countable, \JsonSerializable
         ));
     }
 
+    public function sort(): void
+    {
+        usort($this->violations, static function (Violation $v1, Violation $v2): int {
+            return $v1 <=> $v2;
+        });
+    }
+
     public function jsonSerialize(): array
     {
         return get_object_vars($this);
--- a/tests/E2E/Cli/CheckCommandTest.php
+++ b/tests/E2E/Cli/CheckCommandTest.php
@@ -35,8 +35,8 @@ class CheckCommandTest extends TestCase
         $expectedErrors = 'ERRORS!
 
 App\Controller\Foo has 2 violations
-  should implement ContainerAwareInterface because all controllers should be container aware
   should have a name that matches *Controller because we want uniform naming
+  should implement ContainerAwareInterface because all controllers should be container aware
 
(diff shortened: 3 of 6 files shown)


AlessandroMinoccheri (approved):
LGTM 👍

AlessandroMinoccheri:
@dbu can you please fix the CI?

dbu:
there seems to be a problem with the php 7 builds (phpunit version). i will look at the test fixtures as per the php 8.1 build (i guess the ordering messes with the expected message).

tests/E2E/Cli/CheckCommandTest.php
```diff
@@ -48,8 +48,8 @@ public function test_app_returns_error_with_multiple_violations(): void
   should implement ContainerAwareInterface because all controllers should be container aware
 
 App\Domain\Model has 2 violations
-  depends on App\Services\UserService, but should not depend on classes outside namespace App\Domain because we want protect our domain (on line 14)
-  depends on App\Services\CartService, but should not depend on classes outside namespace App\Domain because we want protect our domain (on line 15)';
+  depends on App\Services\CartService, but should not depend on classes outside namespace App\Domain because we want protect our domain (on line 15)
+  depends on App\Services\UserService, but should not depend on classes outside namespace App\Domain because we want protect our domain (on line 14)';
```
dbu, line 52: this shows something interesting. the `<=>` checks things in order as the properties appear in the class (afaik), so the error message is "more important" than the line number.

src/Rules/Violation.php
```diff
@@ -9,12 +9,12 @@ class Violation implements \JsonSerializable
...
     private $fqcn;
 
-    /** @var string */
-    private $error;
-
     /** @var int|null */
     private $line;
 
+    /** @var string */
+    private $error;
```
dbu, line 16: this is the easiest way to get the sorting with `<=>` right. i think it makes more sense to have fqcn -> line number -> error.

tests/E2E/Cli/CheckCommandTest.php
dbu: i reverted this change and changed the Violation class to have the error after the line number.

dbu:
> there seems to be a problem with the php 7 builds (phpunit version)

i misread the output, that was warnings and skipped tests. the failures were always about the order afaict

src/Rules/Violations.php
```diff
@@ -117,6 +117,13 @@ static function (Violation $a, Violation $b): int {
         ));
     }
 
+    public function sort(): void
```
AlessandroMinoccheri, line 120: could it be useful to add a simple test to document well this new behavior?
Do you think is useful @dbu ?

AlessandroMinoccheri (changes requested):

fain182 (approved):

src/Rules/Violations.php
dbu: yeah, agree that we should have a test for this. i added the test now.

AlessandroMinoccheri (approved):

AlessandroMinoccheri:
Thanks a lot @dbu !

PR #365, opened 2023-03-07
flag to ignore line numbers

dbu:
in addition to #364, this PR adds a flag that allows to ignore line number changes completely. (note that the commit from #364 is also contained in this branch)

when using the flag, the line numbers in the baseline might drift out of sync with reality (but the report uses the line numbers from the actual code). we could potentially not store line numbers in the baseline when the flag is active and instead store that line numbers are ignored, so that the flag is only needed when generating the baseline but not when checking with such a baseline.

there is a drawback when the flag is used: if a specific class has a specific violation, new lines that do the same violation will not be detected anymore, because the ignore is globally.
if we would store the count of a specific violation in the baseline, we could check the count and report if there are more violations than in the baseline. which could also be used for a nicer reporting that groups the violations about the same class into one report, listing on which lines that violation occurs.

how do the maintainers feel about this? i can't promise to work on all this, and for us ignoring the line numbers of the baseline is good enough. but happy to hear your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,19 @@ To avoid using the default baseline file, you can use the `skip-baseline` option
 phparkitect check --skip-baseline
 ```
 
+### Line numbers in baseline
+
+By default, the baseline check also looks at line numbers of known violations.
+When a line before the offending line changes, the line numbers change and the check fails despite the baseline.
+
+With the optional flag `ignore-baseline-linenumbers`, you can ignore the line numbers of violations:
+
+```
+phparkitect check --ignore-baseline-linenumbers
+```
+
+*Warning*: When ignoring line numbers, phparkitect can no longer discover if a rule is violated additional times in the same file.
+
 ## Configuration
 
 Example of configuration file `phparkitect.php`
--- a/src/Rules/Violations.php
+++ b/src/Rules/Violations.php
@@ -106,14 +106,18 @@ class Violations implements \IteratorAggregate, \Countable, \JsonSerializable
         return $this->violations;
     }
 
-    public function remove(self $violations): void
+    /**
+     * @param Violations $violations                Known violations from the baseline
+     * @param bool       $ignoreBaselineLinenumbers If set to true, violations from the baseline are ignored for the same file even if the line number is different
+     */
+    public function remove(self $violations, bool $ignoreBaselineLinenumbers = false): void
     {
+        $comparisonFunction = [__CLASS__, $ignoreBaselineLinenumbers ? 'compareViolationsIgnoreLineNumber' : 'compareViolations'];
+
         $this->violations = array_values(array_udiff(
             $this->violations,
             $violations->violations,
-            static function (Violation $a, Violation $b): int {
-                return $a <=> $b;
-            }
+            $comparisonFunction
         ));
     }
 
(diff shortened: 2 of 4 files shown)


nicolashohm:
I'm really happy to see this PR. I've the same struggle, everytime we change imports or class attributes the line numbers are chaning. Having an option to ignore them would be a good compromise for us too.

Another option could be what psalm is kind doing. They store a code snippet instead of the line number.

dbu:
i just noticed that the baseline stores the full error message, which also contains the list of allowed things. so when the configuration is changed, all baseline entries about classes that have changed rules will not match anymore. i think there is room for further improvements. i guess the second text `but should depend only on classes in one of these namespaces ...` should not be part of the baseline and not be required to match.

i will see if i get to work on that. but would be a separate PR.

i will rebase this PR when #364 is merged, to get the diff reduced to only the line number topic.

AlessandroMinoccheri:
@nicolashohm can you show an example of psalm that stores a code snippet instead of the line number?
I would like to know if can improve Arkitect or if the line number is good enough with this PR also!

Thanks @dbu for your amazing effort and PRs

src/Rules/Violations.php
```diff
@@ -106,17 +106,40 @@ public function toArray(): array
...
-    public function remove(self $violations): void
+    /**
+     * @param Violations $violations Known violations from the baseline
+     * @param bool $ignoreBaselineLinenumbers If set to true, violations from the baseline are ignored for the same file even if the line number is different
+     */
+    public function remove(self $violations, bool $ignoreBaselineLinenumbers = false): void
     {
+        if ($ignoreBaselineLinenumbers) {
+            $comparisonFunction = static function (Violation $a, Violation $b): int {
+                if (($a->getFqcn() === $b->getFqcn()) && ($a->getError() === $b->getError())) {
```
AlessandroMinoccheri, line 117: Can we move this into a specific little function with an explicit name?
I know that it's a clear meaning but what do you think about readability?
Can we improve the readability of this code in your opinion?

nicolashohm:
@AlessandroMinoccheri sure. Let's say there are two `UndefinedMethod` violations saying `The method SomeClass::expects does not exist` in this code:
```php
private SomeClass $mock;
// ...
$this->mock->expects($this->once())
// ...
$this->mock->expects($this->once())
```
The baseline file looks like this:

```xml
<file src="Test.php">
    <UndefinedMethod>
       <code>expects</code>
       <code>expects</code>
    </UndefinedMethod>
</file>
```

It's also not ideal, especially since the code snippet is sometimes really weird/random, like in this example:
`<code>ry\Adapter\O</code>`
But even though it's sometimes bullshit, it works fine for us ✌️

src/Rules/Violations.php
dbu: ah indeed. extracted it.

dbu:
i will have a look at further improving what we do with the baseline, but would propose to still merge this PR to get the  ignore line numbers option.

README.md
```diff
@@ -89,6 +89,17 @@ To avoid using the default baseline file, you can use the `skip-baseline` option
 phparkitect check --skip-baseline
 ```
 
+### Line numbers in baseline
+
+By default, the baseline check also looks at line numbers of known violations.
+When a line before the offending line changes, the line numbers change and the check fails despite the baseline.
+
+With the optional flag `ignore-baseline-linenumbers`, you can ignore the line numbers of violations:
```
micheleorselli, line 97: Could it be worth mentioning the fact that using this option can result in some violations missing?

dbu: good point, yes. i added a warning to the end of this section.

AlessandroMinoccheri (approved):
LGTM 👍

dbu:
@LuigiCardamone not sure if/when i get to it. if you have time please do ;-)

my basic idea was to split Violation::$error into 2 fields: $violation and $context , and only compare $violation (and also only store $violation in the baseline) but when printing the output, we would also print the $context.

AlessandroMinoccheri:
Yes @LuigiCardamone we can discuss about it in a new issue :)

PR #367, opened 2023-03-10
parsing trait also

### LINKED ISSUE #366: Violations on Traits are not reported correctly
### Bug Report

Violations on Traits are not reported correctly.

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Library Version | 0.3.16
| PHP version     | 8.0.28

#### Summary

In my file list I have:

- TraitA.php
- ClassB.php

Overall there are only two violations on `TraitA` but the output of PHPArkitect says "there are 2 violations on ClassB" but that's not true.

This problem is similar to the one that I had before the support for Interfaces was added to PHPArkitect.

AlessandroMinoccheri:
This should fix #366

### DIFF
--- a/src/Analyzer/ClassDescription.php
+++ b/src/Analyzer/ClassDescription.php
@@ -26,6 +26,9 @@ class ClassDescription
     /** @var bool */
     private $interface;
 
+    /** @var bool */
+    private $trait;
+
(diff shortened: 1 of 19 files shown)


fain182 (approved):
lgtm :+1:

PR #370, opened 2023-03-20
don't show progress bar on CI

### LINKED ISSUE #369: PHPArkitect is too verbose in CI
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

```
PHPArkitect 0.3.19

Config file: vendor/mortexa/laravel-arkitect/src/phparkitect.php

analyze class set /home/runner/work/itmap/itmap/app/Http/Controllers

  0/10 [>---------------------------]   0%


analyze class set /home/runner/work/itmap/itmap/app/Exceptions

 10/10 [============================] 100% 0/1 [>---------------------------]   0%


analyze class set /home/runner/work/itmap/itmap/app/Notifications

 1/1 [============================] 100% 0/3 [>---------------------------]   0%


analyze class set /home/runner/work/itmap/itmap/app/Providers

 3/3 [============================] 100% 0/7 [>---------------------------]   0%


analyze class set /home/runner/work/itmap/itmap/app/Policies

 7/7 [============================] 100%  0/12 [>---------------------------]   0%
```

How to disable the progress bar?

fain182:
fixes #369

### DIFF
--- a/src/CLI/Command/Check.php
+++ b/src/CLI/Command/Check.php
@@ -7,11 +7,13 @@ namespace Arkitect\CLI\Command;
 use Arkitect\CLI\Config;
 use Arkitect\CLI\Progress\DebugProgress;
 use Arkitect\CLI\Progress\ProgressBarProgress;
+use Arkitect\CLI\Progress\VoidProgress;
 use Arkitect\CLI\Runner;
 use Arkitect\CLI\TargetPhpVersion;
 use Arkitect\Exceptions\FailOnFirstViolationException;
 use Arkitect\Rules\ParsingErrors;
 use Arkitect\Rules\Violations;
+use OndraM\CiDetector\CiDetector;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -120,7 +122,8 @@ class Check extends Command
             $phpVersion = $input->getOption('target-php-version');
             $targetPhpVersion = TargetPhpVersion::create($phpVersion);
 
-            $progress = $verbose ? new DebugProgress($output) : new ProgressBarProgress($output);
+            $defaultProgressBar = (new CiDetector())->isCiDetected() ? new VoidProgress() : new ProgressBarProgress($output);
+            $progress = $verbose ? new DebugProgress($output) : $defaultProgressBar;
 
             $this->printHeadingLine($output);
 
--- a/src/CLI/Command/Check.php
+++ b/src/CLI/Command/Check.php
@@ -7,13 +7,11 @@ namespace Arkitect\CLI\Command;
 use Arkitect\CLI\Config;
 use Arkitect\CLI\Progress\DebugProgress;
 use Arkitect\CLI\Progress\ProgressBarProgress;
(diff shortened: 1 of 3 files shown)


src/CLI/Command/Check.php
```diff
@@ -120,7 +122,8 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $phpVersion = $input->getOption('target-php-version');
             $targetPhpVersion = TargetPhpVersion::create($phpVersion);
 
-            $progress = $verbose ? new DebugProgress($output) : new ProgressBarProgress($output);
+            $defaultProgressBar = (new CiDetector())->isCiDetected() ? new VoidProgress() : new ProgressBarProgress($output);
+            $progress = $verbose ? new DebugProgress($output) : $defaultProgressBar;
```
AlessandroMinoccheri, line 126: Is it better to move these two lines into a specific factory / class and increase the readability of this code? 
What do you think @fain182 ?

fain182: I tried, but it was worse...

So I simplified this part moving some logic inside the ProgressBarProgress, what do yout think? it's better now or before?

micheleorselli:
Not sure this helps, but why don't we add a new output format suitable for CI environments instead of trying to guess if we are in a ci env? I am thinking to something like a dotted output, like phpunit or phpcsfixer do/ What do you think?

fain182:
> Not sure this helps, but why don't we add a new output format suitable for CI environments instead of trying to guess if we are in a ci env? I am thinking to something like a dotted output, like phpunit or phpcsfixer do/ What do you think?

@micheleorselli I thought about this... It's difficult to give a meaningful output (for example give an idea of the percentage of tests runned) without overwriting the line.
And overwriting the line is exactly what make the ugly output on the CI.

So I think this could be a good solution, I checked the library and supports a lot of different CI, so I suppose it's quite reliable..

micheleorselli:
> > Not sure this helps, but why don't we add a new output format suitable for CI environments instead of trying to guess if we are in a ci env? I am thinking to something like a dotted output, like phpunit or phpcsfixer do/ What do you think?
> 
> @micheleorselli I thought about this... It's difficult to give a meaningful output (for example give an idea of the percentage of tests runned) without overwriting the line. And overwriting the line is exactly what make the ugly output on the CI.
> 
> So I think this could be a good solution, I checked the library and supports a lot of different CI, so I suppose it's quite reliable..

I do not have very strong opinion on that but the less dependencies we have the better it is 
I was wondering how come the output produced by phpunit or phpcsfixer on CI does not have this problem... there should be a way to deal with it without external libraries. if you mind taking a look at them ok, otherwise let's proceed with the external library

dbu:
i would prefer just an explicit flag like `--no-progress` to tell arkitect to be less verbose.

fain182:
@dbu why?

dbu:
its simpler than pulling in a library for this, and when setting up the ci pipeline, it is no problem to add an additional parameter.

fain182:
Adding a library has its cost. But for a change that is so small and localized, the benefit for the user of  having the software that does the "right" thing, is big IMHO.

PR #372, opened 2023-03-21
parsing nullable properties correctly

### LINKED ISSUE #371: Optional Class Property Dependencies not detected
### Bug Report

|    Q            |   A
|---------------- | ------
| Library Version | 0.3.19 
| PHP version     | 8.1

#### Current behavior

When I use optional properties, the dependencies of the properties are not detected.

#### How to reproduce

In FileVisitorTest::test_it_parse_typed_property() make NotBlank optional:
`public ?NotBlank $foo;` 
https://github.com/phparkitect/arkitect/blob/9e5c5a17517d131691c3333ce13dab8a73e817cd/tests/Unit/Analyzer/FileVisitorTest.php#L538

#### Expected behavior

Optional Property Dependencies should be detected correctly.

AlessandroMinoccheri:
It should fix #371

### DIFF
--- a/src/Analyzer/FileVisitor.php
+++ b/src/Analyzer/FileVisitor.php
@@ -5,6 +5,7 @@ namespace Arkitect\Analyzer;
 
 use PhpParser\Comment\Doc;
 use PhpParser\Node;
+use PhpParser\Node\NullableType;
 use PhpParser\NodeVisitorAbstract;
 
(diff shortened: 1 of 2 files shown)


fain182 (approved):
It's ok, I think in the future we could decorate the $node because it doesn't represents what we need... for example $node->type now it does not rappresent the type that we are interested in.

PR #386, opened 2023-05-25
adds fix to consider valid a single letter class

### LINKED ISSUE #385: Single letter class names not permitted
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | No
| Library Version | `0.3.25`
| PHP version     | `8.0.28`

#### Summary

We would love to adopt PHPArkitect to enforce some guidelines in a rather large monolith. We ran into an issue running `check` against our codebase due to the error `\T is not a valid namespace definition`.

Traced down to the file `FullyQualifiedClassName`:
```php
public static function fromString(string $fqcn): self
    {
        $validFqcn = '/^[a-zA-Z_\x7f-\xff\\\\][a-zA-Z0-9_\x7f-\xff\\\\]*[a-zA-Z0-9_\x7f-\xff]$/';

        if (!(bool) preg_match($validFqcn, $fqcn)) {
            throw new \RuntimeException("$fqcn is not a valid namespace definition");
        }

        // ...
    }
```

The issue is that we have a class `\T` which is used more than 9,000 times in the codebase but the regex defined in `\Arkitect\Analyzer\FullyQualifiedClassName::fromString` stipulates that classes **must** be 2 character long to be valid.

P.S. Context for this class `\T`... It is the only instance of a single letter class that we have and its purpose is for tagging text for be translated. It's an old part of the codebase but used throughout our project.

#### Current behavior

Class `\T` is not determined to be a valid FQCN. PHP Arkitect throws an exception and does not analyse the file.

#### How to reproduce

Create a file with FQCN with only a single letter.
```php
class T {
    public static function _(string $text): {}
}
```

Run PHPArkitect over the directory with the class in it.
```sh
vendor/bin/phparkitect check
```

Result:
```
Parse Error: T is not a valid namespace definition
```

#### Expected behavior

`\T` should be recognised as a valid FQCN by `FullyQualifiedClassName::fromString`.

This is probably debatable however I will provide our reasoning for permitting this...
- This is a valid PHP class according to the [PHP Manual - Class](https://www.php.net/manual/en/language.oop5.basic.php#:~:text=As%20a%20regular%20expression%2C%20it%20would%20be%20expressed%20thus%3A)
- Refactoring to rename this class would probably be risky and/or not worth the time to "fix". I can't imagine that we'd be the only organisation who will encounter this as a blocker for adopting the tool.
- This will enable us to adopt PHPArkitect within a large and active PHP codebase with a mix of old & new styles of PHP code. If successful it would be likely for us to contribute PR's as/when we come across use/edge-cases.

fain182:
fixes #385 

We loose the check that a class should not be start with a number, but I don't think is critical...

### DIFF
--- a/src/Analyzer/FullyQualifiedClassName.php
+++ b/src/Analyzer/FullyQualifiedClassName.php
@@ -58,7 +58,7 @@ class FullyQualifiedClassName
 
     public static function fromString(string $fqcn): self
     {
-        $validFqcn = '/^[a-zA-Z_\x7f-\xff\\\\][a-zA-Z0-9_\x7f-\xff\\\\]*[a-zA-Z0-9_\x7f-\xff]$/';
+        $validFqcn = '/^[a-zA-Z0-9_\x7f-\xff\\\\]*[a-zA-Z0-9_\x7f-\xff]$/';
 
         if (!(bool) preg_match($validFqcn, $fqcn)) {
             throw new \RuntimeException("$fqcn is not a valid namespace definition");
--- a/src/Analyzer/NameResolver.php
+++ b/src/Analyzer/NameResolver.php
@@ -365,8 +365,6 @@ class NameResolver extends NodeVisitorAbstract
      * @psalm-suppress MissingParamType
      * @psalm-suppress PossiblyNullArgument
      * @psalm-suppress MissingReturnType
-     *
-     * @param mixed $node
      */
     private function resolveType($node)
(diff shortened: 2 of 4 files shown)


fain182:
I have not more cs-fix errors on my laptop with php 8.2, but I see there are errors on the CI... 
Maybe older PHP versions use older version of php-cs-fix that have a different behavior?

PR #387, opened 2023-05-28
Implement logical or expression

akondas:
## Motivation

Allow to define logical expressions which will help with more advanced or complex expressions. Class name is not perfect but `Or` is keyword hence can't be used directly as class name. `Orx` was inspired by doctrine project.

If we are fine with that approach then I can implement more (like `and`).

### DIFF
--- a/src/Expression/Boolean/Orx.php
+++ b/src/Expression/Boolean/Orx.php
@@ -1,0 +1,43 @@
+<?php
+declare(strict_types=1);
+
+namespace Arkitect\Expression\Boolean;
+
+use Arkitect\Analyzer\ClassDescription;
+use Arkitect\Expression\Description;
+use Arkitect\Expression\Expression;
+use Arkitect\Rules\Violation;
+use Arkitect\Rules\ViolationMessage;
+use Arkitect\Rules\Violations;
+
+final class Orx implements Expression
+{
+    /** @var Expression[] */
+    private $expressions;
+
+    public function __construct(array $expressions)
+    {
+        $this->expressions = $expressions;
+    }
+
+    public function describe(ClassDescription $theClass, string $because): Description
+    {
+        return new Description('at least one expression must be true', $because);
+    }
+
+    public function evaluate(ClassDescription $theClass, Violations $violations, string $because): void
+    {
+        foreach ($this->expressions as $expression) {
+            $newViolations = new Violations();
+            $expression->evaluate($theClass, $newViolations, $because);
+            if (0 === $newViolations->count()) {
+                return;
+            }
+        }
+
+        $violations->add(Violation::create(
+            $theClass->getFQCN(),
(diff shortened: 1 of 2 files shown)


src/Expression/Boolean/Orx.php
```diff
@@ -0,0 +1,43 @@
...
+
+final class Orx implements Expression
+{
+    /** @var Expression[] */
+    private $expressions;
+
+    public function __construct(array $expressions)
+    {
+        $this->expressions = $expressions;
+    }
```
noelma, line 21: **suggestion**: use variable length argument lists.
```suggestion
    public function __construct(...$expressions)
    {
        $this->expressions = $expressions;
    }
```
```php
new Orx(new Extend('My\BaseClass'), new Extend('Your\OtherClass'));
```
because in phparkiteck upgrade in ^7.4 it will make typing easier:
```suggestion
    public function __construct(Expression ...$expressions)
    {
        $this->expressions = $expressions;
    }
```
It will still be possible to use it with an array such as :
```php
new Orx(...[new Extend('My\BaseClass'), new Extend('Your\OtherClass')]);
```

src/Expression/Boolean/Orx.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function describe(ClassDescription $theClass, string $because): Description
+    {
+        return new Description('at least one expression must be true', $because);
+    }
+
+    public function evaluate(ClassDescription $theClass, Violations $violations, string $because): void
+    {
+        foreach ($this->expressions as $expression) {
```
noelma, line 30: **question**: Could we throw a logicalException or violation if none or only one expression is used? And if so, add a test method :pray:

akondas: okay :vulcan_salute:

akondas: added with tests

src/Expression/Boolean/Orx.php
```diff
@@ -0,0 +1,46 @@
...
+use Arkitect\Rules\Violation;
+use Arkitect\Rules\ViolationMessage;
+use Arkitect\Rules\Violations;
+
+final class Orx implements Expression
+{
+    /** @var Expression[] */
+    private $expressions;
+
+    public function __construct(array $expressions)
```
hgraca, line 18: Could we do `public function __construct(Expression ...$expressions)`?

src/Expression/Boolean/Orx.php
hgraca: I'm actually not sure this is a good idea.
If we want to build automations on top of this, we will want to instantiate this with a list of expressions that are somehow inferred.
This list might have something or might be empty, in which case I think it should be irrelevant and therefore should not fail, essentially being ignored.

WDYT?

PR #390, opened 2023-06-19
Bug: Anonymous class extends is parsed instead of class extends

jdomenechbLendable:
An anonymous class declared inside a normal class makes the parser take the `extends` from the anonymous class, instead of the correct one.

```
<?php
namespace Root\Animals;

class Animal
{
}

class Cat extends Animal
{
    public function methodWithAnonymous(): void
    {
        $obj = new class extends \stdClass {};
    }
}
```
This code above takes `stdClass` instead of the expected `Animal`.


felipefrancisco:
+1

AlessandroMinoccheri:
@jdomenechbLendable can you fix coding standards?

jdomenechbLendable:
@AlessandroMinoccheri Done. Imported main from original repository. Local build tells me no issues to be solved.

PR #391, opened 2023-06-19
Bug: Class ends pattern matches classes that contain the pattern

jdomenechbLendable:
This is due to not checking strictly that the matched name must end with the pattern.


micheleorselli:
This should have been fixed as confirmed by the test added in #471, let us know if we can close it

PR #394, opened 2023-07-26
Create a Not expression

hgraca:
This will reduce the amount of expressions needed
and allow for complex expressions (within an OR
or an AND) to be negated.

### DIFF
--- a/src/Expression/Boolean/Not.php
+++ b/src/Expression/Boolean/Not.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Arkitect\Expression\Boolean;
+
+use Arkitect\Analyzer\ClassDescription;
+use Arkitect\Expression\Description;
+use Arkitect\Expression\Expression;
+use Arkitect\Rules\Violation;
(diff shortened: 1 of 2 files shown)


fain182:
We already tried this, it would be nice because we could remove some rules, but the error messages are much less readable.
I think we should prioritize DX over our code simplicity...

hgraca:
I understand, however, it's not just your code, it's also the code of the developers using this library.

1. There are rules that are specific to a project, and not really useful anywhere else, and developers could build just the positive version and negate it using the `Not` provided by the library;

2. There are complex and dynamic rules that we can't build if we don't have the `Not`, for example, [the PR that you liked](https://github.com/phparkitect/arkitect/pull/399) is not possible, or at least not elegantly, done without this.

3. From what I understand, your reservation is not about the rule, but about the feedback when it fails, so I would say it's a different issue to tackle, maybe for boolean expressions we should have a different way of composing the error message? I feel that `must NOT (HappyIsland should be an interface) because we want to add this rule for our software` is acceptable feedback for a developer to read as a tradeoff for using this expression, and it doesn't stop anyone from building specific `NotFooBar` expressions.

PR #395, opened 2023-07-26
Create a And expression

hgraca:
This will allow for complex nested expressions, like ANDs inside ORs.


fain182:
Arkitekt already supports `AndThat()` (but we forgot to document it, sorry), do you have use cases where this should not be enough?

hgraca:
Yes, `andThat` only works at the root of the expression, If we have a `Not` or an `Or` and inside we want to have an `And`, atm we can't do it.

PR #396, opened 2023-07-26
Create a IsA expression

hgraca:
This will allow for testing if a class extends or
implements another code unit, anywhere in the inheritance tree.

### DIFF
--- a/src/Expression/ForClasses/IsA.php
+++ b/src/Expression/ForClasses/IsA.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Arkitect\Expression\ForClasses;
+
+use Arkitect\Analyzer\ClassDescription;
+use Arkitect\Expression\Description;
+use Arkitect\Expression\Expression;
+use Arkitect\Rules\Violation;
+use Arkitect\Rules\ViolationMessage;
+use Arkitect\Rules\Violations;
+
+final class IsA implements Expression
+{
+    /** @var array<class-string> */
+    private array $allowedFqcnList;
+
+    /**
+     * @param class-string ...$allowedFqcnList
+     */
+    public function __construct(string ...$allowedFqcnList)
+    {
+        $this->allowedFqcnList = $allowedFqcnList;
+    }
+
+    public function describe(ClassDescription $theClass, string $because = ''): Description
+    {
+        $allowedFqcnList = implode(', ', $this->allowedFqcnList);
+
+        return new Description("should inherit from one of: $allowedFqcnList", $because);
+    }
+
+    public function evaluate(ClassDescription $theClass, Violations $violations, string $because = ''): void
+    {
+        if (!$this->isA($theClass, ...$this->allowedFqcnList)) {
+            $violation = Violation::create(
+                $theClass->getFQCN(),
+                ViolationMessage::selfExplanatory($this->describe($theClass, $because)),
+                $theClass->getFilePath()
+            );
+
+            $violations->add($violation);
+        }
+    }
+
+    /**
+     * @param class-string ...$allowedFqcnList
+     */
+    private function isA(ClassDescription $theClass, string ...$allowedFqcnList): bool
+    {
+        foreach ($allowedFqcnList as $allowedFqcn) {
+            if (is_a($theClass->getFQCN(), $allowedFqcn, true)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+}
--- a/src/Expression/ForClasses/IsA.php
+++ b/src/Expression/ForClasses/IsA.php
@@ -13,27 +13,25 @@ use Arkitect\Rules\Violations;
 
 final class IsA implements Expression
 {
-    /** @var array<class-string> */
-    private array $allowedFqcnList;
+    /** @var class-string */
+    private string $allowedFqcn;
 
     /**
-     * @param class-string ...$allowedFqcnList
+     * @param class-string $allowedFqcn
      */
-    public function __construct(string ...$allowedFqcnList)
+    public function __construct(string $allowedFqcn)
     {
-        $this->allowedFqcnList = $allowedFqcnList;
+        $this->allowedFqcn = $allowedFqcn;
     }
 
     public function describe(ClassDescription $theClass, string $because = ''): Description
     {
-        $allowedFqcnList = implode(', ', $this->allowedFqcnList);
-
-        return new Description("should inherit from one of: $allowedFqcnList", $because);
+        return new Description("should inherit from: $this->allowedFqcn", $because);
     }
 
     public function evaluate(ClassDescription $theClass, Violations $violations, string $because = ''): void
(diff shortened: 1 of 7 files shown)


src/Expression/ForClasses/IsA.php
```diff
@@ -0,0 +1,56 @@
...
+            $violations->add($violation);
+        }
+    }
+
+    private function isA(ClassDescription $theClass, string ...$allowedFqcnList): bool
+    {
+        foreach ($allowedFqcnList as $allowedFqcn) {
+            if (
+                \is_a($theClass->getFQCN(), $allowedFqcn, true)
+                || \is_subclass_of($theClass->getFQCN(), $allowedFqcn)
```
jdreesen, line 48: Do we really need to check both? Both do the same, except that `is_a()` also returns `true` for the class itself (like `is_a(A::class, A::class)`), AFAIK, so `is_subclass_of()` should never be called.

jdreesen: > unfortunately, `is_a` doesn't check for interface implementation

I think it does: https://3v4l.org/91Yt8

AFAIK, `is_a` does the same as `instanceof` (and it is even more powerful, since it can check string class names and not only instances, if the third parameter is `true`).

hgraca: HAH! u r correct, I will fix this

hgraca: fixed and force pushed

jdreesen (approved):

micheleorselli:
> Hello, this PR is useful. When can it merge?

When we used to support older php versions we wanted to avoid relying on checks triggering the autoload and parse the code statically. Not sure it still makes sense, so potentially this is a good addition. At this point I am wondering if we should just update `Extend/NotExtend` `Implement/NotImplement`. wdyt @AlessandroMinoccheri @fain182?

hgraca:
@micheleorselli 
> avoid relying on checks triggering the autoload 

That is why I opened [this other PR](https://github.com/phparkitect/arkitect/pull/401) but never got it merged. 
This would effectively put the decision on the user; if they wanted to use an expression that would need to trigger the autoload, they would be able to, which is not the case at the moment.

That PR just adds an IF statement so we don't try to define PHPARKITECT_COMPOSER_INSTALL more than once, which triggers PHP errors.
I don't see a negative point in merging it, but if it is not desirable, let me know so I can close that PR.

AlessandroMinoccheri:
For me, the PR is adding value.
What is your idea of extending `Extend/NotExtend` and `Implement/NotImplement` @micheleorselli ?

Maybe we can consider both ways to understand what is the correct one.

micheleorselli:
As of now, there is a known limitation when we parse files building ClassDescription objects as we do not walk the inheritance tree. This means that, given a ClassDescription object, the `extends` and `implement` attributes _contains only classes that are **directly** extended or implemented_.  

If I have a rule saying "class X should not extend class Y" is that the expected behaviour? I doubt that 🤔 

One possibile approach would be having to set of rules like
- `Extends`, which work using is_a and triggering autoload
- `ExtendsDirectly` which behaves like the current Extends

this also would work for `implements` for interfaces, `use` for traits, and so on  

wdty? @AlessandroMinoccheri @fain182

fain182:
> When we used to support older php versions we wanted to avoid relying on checks triggering the autoload and parse the code statically.

I think we need to ensure that autoloading works with all our installation methods (Composer, PHAR, Docker); otherwise, it's a good idea. (I'm even open to deprecating some of the installation methods.)

hgraca:
@micheleorselli I made some necessary updates to this PR, it all seems to be green, please feel free to squash&merge or ask for changes.

src/Expression/ForClasses/IsA.php
```diff
@@ -0,0 +1,60 @@
...
+    private function isA(ClassDescription $theClass, string ...$allowedFqcnList): bool
+    {
+        foreach ($allowedFqcnList as $allowedFqcn) {
+            if (is_a($theClass->getFQCN(), $allowedFqcn, true)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
```
micheleorselli, line 59: If I got it right here we need the class to implement at least one of the fqcn specified. I'm ok with that but I was wondering if we can make it explicit in some way. For instance we could either: 
- make IsA accepting one parameter instead of an array (but it is less expressive since you cannot say "a class is A or B")
- change the name to something like `IsOneOf` or something similar

tests/Unit/Expressions/ForClasses/DummyClasses/Banana.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Arkitect\Tests\Unit\Expressions\ForClasses\DummyClasses;
+
+class Banana implements FruitInterface
```
micheleorselli, line 7: if possibile I would avoid creating test classes here. You could either:

1.  create fixtures directory like we did in the E2E section of the tests
2.  rely on vfsstream and create the classes only in memory. This also means you'll have to create and register a custom autoloader. (that would be my preferred choice but if it is too much work no 1. is fine

micheleorselli (changes requested):

src/Expression/ForClasses/IsA.php
hgraca: I changed the name to `IsOneOf`

tests/Unit/Expressions/ForClasses/DummyClasses/Banana.php
hgraca: I moved the dummy classes declaration to inside the class test itself, this avoids creating the dummy classes files, keeps the whole test self-contained and avoids the complexity of using a vfs.

Let me know if this is ok.

hgraca:
@micheleorselli I squashed the previous fixup commits and created 2 more, one for each comment.

let me know if anything else is needed from my side.

src/Expression/ForClasses/IsOneOf.php
```diff
@@ -0,0 +1,60 @@
...
+
+final class IsOneOf implements Expression
+{
+    /** @var array<class-string> */
+    private array $allowedFqcnList;
+
+    /**
+     * @param class-string ...$allowedFqcnList
+     */
+    public function __construct(string ...$allowedFqcnList)
```
AlessandroMinoccheri, line 22: Can we avoid variadic parameter?
We removed it from other expression to add the exclude parameter if needed

hgraca: Sure.
I dropped the commit that renamed the expression, and applied the previous suggestion where we only accept one fqcn to test, instead of the variadic.

let me know if anything else is needed from my side.

AlessandroMinoccheri:
What do you think @micheleorselli? For me we can merge it.

micheleorselli:
yup, all good for me. Thank you @hgraca 🎉

PR #397, opened 2023-07-26
Create a HasCorrespondingUnit expression

hgraca:
This will allow us to ensure that certain classes always have a test, or that every test has a matching class and their namespaces are correct.


hgraca:
I converted this back to draft cozz I found some edge cases that are not working.
Once I fix that I will put the PR back to ready.

hgraca:
This is actually fine, my issue was at another place :)

hgraca:
@AlessandroMinoccheri done! The pipelines seem to not be running atm though...

lzakrzewski (approved):

marmichalski:
@AlessandroMinoccheri @micheleorselli is there anything left to be done here?

PR #398, opened 2023-07-27
Create a IsMapped expression

hgraca:
This will allow us to ensure that certain classes are always in a map. This is useful, for example, to ensure that DTOs like commands and events are always set in a map, so that a serializer knows how to serialize/deserialize them.


AlessandroMinoccheri:
@hgraca  Before approving, can you please add something to the readme file :) ?

PR #399, opened 2023-07-27
Allow architectures to be created from expressions

hgraca:
This will allow for composable and more complex rules.

DEPENDS ON [#394](https://github.com/phparkitect/arkitect/pull/394) [#395](https://github.com/phparkitect/arkitect/pull/395)

Only the last commit is relevant for this PR.


fain182:
That's really cool, it is much more coherent with the rest of design than the way we are doing it now...

hgraca:
@fain182 the issues should be fixed (cant tell cozz the build isn't running) can you check if there's something more to do pls?

PR #400, opened 2023-07-27
Multiple dirs class set

hgraca:
Currently, if we want to include two folders from the project root, we need to add the root of the project and then exclude everything we dont want to include, and remember to add a folder to the config if we add a new folder to the project root.

With the `MultipleDirClassSet` we can specify just the directories we want it to scan,
removing all cruft from our config.

### DIFF
--- a/src/ClassSet.php
+++ b/src/ClassSet.php
@@ -11,13 +11,14 @@ use Symfony\Component\Finder\Finder;
  */
 class ClassSet implements \IteratorAggregate
 {
-    private string $directory;
+    /** @var string[] */
+    private array $directoryList;
 
     private array $exclude;
 
-    private function __construct(string $directory)
+    private function __construct(string ...$directoryList)
     {
-        $this->directory = $directory;
+        $this->directoryList = $directoryList;
         $this->exclude = [];
     }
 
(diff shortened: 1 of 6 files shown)


jdreesen:
How about refactoring `ClassSet` so that it can handle multiple directories instead? 

The constructor is `private` so we could change it (without BC break) and add a second static constructor `fromDirs(string ...$directoryList)`. It shouldn't be a breaking change, right?

hgraca:
Yes! That was my first approach, but I found some reason to do it differently... Dont remember what it was though. :(

We can even change the named constructor without a BC break.

my only concern here is with the `getDir()` method, cozz its return will not be atomic anymore, but on the other hand we are only using it for writing it on the screen.

Anyway, as I can't see anymore what drove me to this approach, I will simplify as you suggest.

hgraca:
@jdreesen done, and force pushed

src/ClassSet.php
```diff
@@ -7,15 +7,15 @@
 
 class ClassSet implements \IteratorAggregate
 {
-    /** @var string */
-    private $directory;
+    /** @var string[] */
+    private $directoryList;
```
AlessandroMinoccheri, line 11: I was thinking of a var name like `directories` instead of `directoryList`, what do you think?

hgraca: It does read better, however when plural prefer to suffix with "List", otherwise in most cases the difference between plural and singular will be only a letter (s), wich can easily be overlooked and result in confusion.

That said, if my argument doesn't convince you let me know and I will gladly change it. :)

And tkx for the CR btw.

AlessandroMinoccheri:
@hgraca Before approving, can you please add something to the readme file :) ?

hgraca:
@AlessandroMinoccheri done

marmichalski (approved):

hgraca:
@micheleorselli @AlessandroMinoccheri 

I rebased, fixed the conflicts, and added 2 commits.

Let me know if this is ok or if I can do something else.

micheleorselli:
Looks good to me, thank you @hgraca! 🙇

PR #401, opened 2023-07-29
Don't assume PHPARKITECT_COMPOSER_INSTALL is not defined yet

hgraca:
By default, the classmap is not loaded, so it's not possible to use functions like `is_a` to infer inheritance of classes under test. To solve this, the autoloader must be included in the config file, but then it will try to define this variable again, which will cause it to throw an error.

### DIFF
--- a/bin-stub/phparkitect
+++ b/bin-stub/phparkitect
@@ -4,11 +4,13 @@ declare(strict_types=1);
 
 use Arkitect\CLI\PhpArkitectApplication;
 
-foreach (array(__DIR__ . '/../../../autoload.php', __DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php') as $file) {
-    if (file_exists($file)) {
-        define('PHPARKITECT_COMPOSER_INSTALL', $file);
+if (!defined('PHPARKITECT_COMPOSER_INSTALL')) {
+    foreach (array(__DIR__ . '/../../../autoload.php', __DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php') as $file) {
+        if (file_exists($file)) {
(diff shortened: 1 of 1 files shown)


fain182:
Can you show a snippet of how you use `is_a` in the configuration file to better understand your needs?

hgraca:
Like this:

```php
<?php

declare(strict_types=1);

use Arkitect\ClassSet;
use Arkitect\CLI\Config;
use Arkitect\Expression\ForClasses\HaveNameMatching;
use Arkitect\Expression\ForClasses\IsA;
use Arkitect\Rules\Rule;
use Hgraca\SymfonyBoilerplate\Infrastructure\CommandBus\Port\CommandHandlerInterface;

require_once dirname(__DIR__) . '/../vendor/autoload_runtime.php';

return static function (Config $config): void {
    $rootPath = realpath(__DIR__ . '/../..');

    $classSet = ClassSet::fromDir("$rootPath/src", "$rootPath/tests");

    $rules[] = Rule::allClasses()
        ->except(...$this->exceptionList)
        ->that(new IsA(CommandHandlerInterface::class))
        ->should(new HaveNameMatching('*CommandHandler'))
        ->because("our command handlers must be suffixed 'CommandHandler', to differentiate from 'QueryHandler' and 'MutatorHandler'.");

    $config->add($classSet, ...$rules);
};
```
I need to use the `IsA` expression there, as opposed to `Implements`, because `Implements` only checks for direct implementation, not implementation through the inheritance tree.

micheleorselli:
@hgraca release [0.5.4](https://github.com/phparkitect/arkitect/releases/tag/0.5.4) allows to specify an extra autoloader as an option if you'd like to give it a try

hgraca:
@micheleorselli I rebased my fork, removed my hack, tested it in my current project from work and it works like a charm, as far as I can see.

tkx a lot! 

I am therefore closing this PR.

PR #402, opened 2023-07-29
Allow to build rules excluding by expression

hgraca:
This will allow us to exclude classes using complex expressions, instead of just by FQCN matching.

### DIFF
--- a/src/Analyzer/ClassDescription.php
+++ b/src/Analyzer/ClassDescription.php
@@ -3,6 +3,9 @@ declare(strict_types=1);
 
 namespace Arkitect\Analyzer;
 
+use Arkitect\Expression\Expression;
+use Arkitect\Rules\Violations;
+
 class ClassDescription
 {
     /** @var FullyQualifiedClassName */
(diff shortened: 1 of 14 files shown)


fain182:
Hi Herberto, thanks for you contributions you brought in a lot of interesting PR :slightly_smiling_face: 

I am not sure about the exceptExpression, because using, you example:
```
Rule::allClasses()
            ->exceptExpression(new HaveNameMatching('*TestCase'))
            ->that(new ResideInOneOfTheseNamespaces('Arkitect\Tests\Fixtures\Animal'))
```
Can be already written as
```
Rule::allClasses()
            ->that(new ResideInOneOfTheseNamespaces('Arkitect\Tests\Fixtures\Animal'))
            ->andThat(new NotHaveNameMatching('*TestCase'))
```

But now that I check I saw that we forgot to document the `AndThat` in the README... maybe we should improve this.

hgraca:
Hah, I see, so `exceptExpression` is actually redundant, except for the semantics of `Not` being included in the method so we can use the positive version of the expression.

Even the current `except` method is redundant cozz

```php
Rule::allClasses()
            ->except('*TestCase')
            ->that(new ResideInOneOfTheseNamespaces('Arkitect\Tests\Fixtures\Animal'))
```
can also be written as
```php
Rule::allClasses()
            ->that(new ResideInOneOfTheseNamespaces('Arkitect\Tests\Fixtures\Animal'))
            ->andThat(new NotHaveNameMatching('*TestCase'))
```

Personally, I like the easiness of just using a string to say what classes to exclude, but I would make the `Expression`s the main actor in this library, in the sense that every method should be able to accept at least an expression, and maybe a string for simplification.

I do see your point though, feel free to close the PR if you feel this is not the way to go, no hard feelings. :)

fain182:
The except is intended to contains a blacklist of class that for some reason you don't want to check. So even if i understand your idea of coherency, I think in this case a collection of FQCN may be the best option.

I close this PR for now :slightly_smiling_face:

PR #403, opened 2023-07-29
[BC_BREAK] Replace `array` for `string ...`

hgraca:
This provides better typing and ease of use


tests/Unit/Analyzer/ClassDescriptionTest.php
```diff
@@ -36,7 +36,7 @@ public function test_should_return_true_if_there_class_is_in_namespace_array():
     {
         $cd = $this->builder->build();
 
-        $this->assertTrue($cd->namespaceMatchesOneOfTheseNamespaces(['Fruit']));
+        $this->assertTrue($cd->namespaceMatchesOneOfTheseNamespaces('Fruit'));
```
AlessandroMinoccheri, line 39: I would suggest adding at least another string to show that we can add more arguments

AlessandroMinoccheri (changes requested):

tests/Unit/Analyzer/ClassDescriptionTest.php
AlessandroMinoccheri: @hgraca Can we continue to support array and the new behavior to not create a BC?

hgraca: > I would suggest adding at least another string to show that we can add more arguments

I added another string to the test case.

> Can we continue to support array and the new behavior to not create a BC?

Do you have a suggestion? Or should I add a new method? (but with what name?)
Also, since the library is not stable yet, I think ppl using this should expect frequent BC breaks and have their dependencies setup accordingly (ie `^0.2` will not automatically update to `0.3`, unlike with stable versions)

AlessandroMinoccheri: Yeah we can introduce BC but I would like t know if we can maintain a retro-compatibility with this feature :)

hgraca: I added a fixup command adding a new method instead of changing the current method signature.

This avoids the BC break. We can also deprecate the old method and later recreate it with the new signature, for the sake of having a nice method name.

PR #404, opened 2023-08-07
Fix pattern match

hgraca:
At some point this was correct and then changed into a bug.

I am reverting the change that introduced the bug.

### DIFF
--- a/src/Expression/ForClasses/ResideInOneOfTheseNamespaces.php
+++ b/src/Expression/ForClasses/ResideInOneOfTheseNamespaces.php
@@ -32,7 +32,7 @@ class ResideInOneOfTheseNamespaces implements Expression
     {
         $resideInNamespace = false;
         foreach ($this->namespaces as $namespace) {
-            if ($theClass->namespaceMatches($namespace)) {
+            if ($theClass->namespaceMatches($namespace.'*')) {
                 $resideInNamespace = true;
             }
         }
--- a/src/Analyzer/PatternString.php
+++ b/src/Analyzer/PatternString.php
@@ -27,7 +27,7 @@ class PatternString
             return $isInThisNamespace || $isThisClass;
         }
 
-        return $this->startsWithPattern($pattern);
+        return fnmatch($pattern, $this->value, \FNM_NOESCAPE);
     }
 
     public function toString(): string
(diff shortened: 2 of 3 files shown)


fain182:
Thanks for you contributions! As you can see from the broken tests this function is used in a lot of places, and reverting to an old behavior is not a valid solution... 
Which is the exact behavior that you didn't expect? If you have an example would be great :slightly_smiling_face:

hgraca:
@fain182 

Reverting the changes and leaving the test as is, we can see that this assertion fails:

```php
        $pattern = new PatternString('SoThisIsAnExample');
        $this->assertFalse($pattern->matches('*This'));
```

However, it shouldn't fail because it shouldn't match since there is no wildcard at the end of the string.

hgraca:
@fain182 
I removed a duplicate assertion and added a few more to make sure all scenarios are covered.

hgraca:
@fain182 I believe I fixed the issues, can you check the last commits, please?

d-cichon:
@hgraca i need this ;-;

Can we merge it?

hgraca:
@d-cichon as far as I'm concerned, yes. However, I'm not a maintainer, I can't merge anything. 
Back when I opened these PRs I ended up making my own fork cozz I felt this project was abandoned, no one was merging anything.
Not sure how this is now, but u might need to do the same.

hgraca:
@d-cichon @fain182 

I see there is work being merged, so I guess this is not abandoned as I thought.

Thus, I opened an [issue](https://github.com/phparkitect/arkitect/issues/457) for this PR, rebased, squashed and force pushed to this branch.

This PR resolves https://github.com/phparkitect/arkitect/issues/457

AlessandroMinoccheri (approved):
LGTM 👍

AlessandroMinoccheri:
@fain182 @micheleorselli  it seems OK for me, any other opinions?

src/Expression/ForClasses/ResideInOneOfTheseNamespaces.php
```diff
@@ -32,7 +32,7 @@ public function evaluate(ClassDescription $theClass, Violations $violations, str
     {
         $resideInNamespace = false;
         foreach ($this->namespaces as $namespace) {
-            if ($theClass->namespaceMatches($namespace)) {
+            if ($theClass->namespaceMatches($namespace.'*')) {
```
micheleorselli, line 35: just out of curiosity, why do we need to add this?

hgraca: Its from a long time ago, but I guess it was because if a class is in a subnamespace is still in the given namespace.

However, looking at it now, it's out of scope fo this PR, and not having the `*` is a better option because we can match the namespace exactly or provide the namespace with a `*` which will match subnamespaces as well.

So, Im adding a fixup commit to remove this change.

hgraca:
I rebased and force pushed

PR #408, opened 2023-09-18
Simplify usage of FileParserFactory

hgraca:
By allowing createFileParser to be called with null we need less code everywhere else to create a default file parser.

### DIFF
--- a/src/Analyzer/FileParserFactory.php
+++ b/src/Analyzer/FileParserFactory.php
@@ -9,13 +9,15 @@ use PhpParser\NodeTraverser;
 
 class FileParserFactory
 {
-    public static function createFileParser(TargetPhpVersion $targetPhpVersion, bool $parseCustomAnnotations = true): FileParser
-    {
+    public static function createFileParser(
+        TargetPhpVersion $targetPhpVersion = null,
+        bool $parseCustomAnnotations = true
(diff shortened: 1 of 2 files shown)


src/Analyzer/FileParserFactory.php
```diff
@@ -9,13 +9,15 @@
 
 class FileParserFactory
 {
-    public static function createFileParser(TargetPhpVersion $targetPhpVersion, bool $parseCustomAnnotations = true): FileParser
-    {
+    public static function createFileParser(
+        TargetPhpVersion $targetPhpVersion = null,
```
marmichalski, line 13:
```suggestion
        ?TargetPhpVersion $targetPhpVersion = null,
```

I assume this should be explicitly marked as a nullable type instead of depending on implicit PHP behaviour.

src/CLI/TargetPhpVersion.php
```diff
@@ -26,7 +26,7 @@ private function __construct(?string $version)
         $this->version = $version;
     }
 
-    public static function create(?string $version): self
+    public static function create(string $version = null): self
```
marmichalski, line 29:
```suggestion
    public static function create(?string $version = null): self
```

as above

PR #409, opened 2023-09-18
Prevent duplicate namespaces in ResideInOneOfTheseNamespaces

hgraca:
When building expressions automatically, with a builder, it sometimes happens that a namespace is added more than once.

Besides being redundant to have and check the same namespace more than once, the tests also become more difficult to assert on their outcomes.


AlessandroMinoccheri:
@hgraca is it possible to add a test for this behavior?

hgraca:
I'm sorry man, but it's been a long time, I had to fork this project and now I don't have time for this anymore.
Feel free to close or take over my PRs.
Maybe at some point I will have time to try to contribute again, but not in the foreseeable future.
Best of luck.

marmichalski:
Happy to continue some of @hgraca legacy: https://github.com/phparkitect/arkitect/pull/431

PR #411, opened 2023-09-18
Add $because to rules builder

hgraca:
This will allow the output to be more expressive


marmichalski (approved):
LGTM might want to update (or add a new) one of the ArchitectureTest to verify that the argument string is indeed being used :+1:

AlessandroMinoccheri (approved):
LGTM

PR #413, opened 2023-10-26
Fix documentation about class NotResideInTheseNamespaces

LuigiCardamone:
This PR fix an error in the README file. In the README the class `NotResideInOneOfTheseNamespaces` does not exist. The correct one is `NotResideInTheseNamespaces`.


fain182:
Maybe we should rename for consistency, but in the meantime fix the readme is a great! thanks :+1:

PR #421, opened 2024-01-16
Fix reference consistency in `NameResolver` to resolve CI issues

ben-challis:
I initially just ran `php-cs-fixer` to address this however that resulted in inconsistent references within the file, so I've gone with manually fixing it to the most consistent usages, long format for `Use_` and short format for `Name`.


vibby:
Thank you @ben-challis ! Can you review and merge, please @fain182 ?

PR #422, opened 2024-01-17
feature: Add IsReadonly Expression

helyakin:
Hey there !

First of all I wanted to say how cool I find this package. It is very useful so thank you everyone !

I've started coding before reading the guidelines so maybe this will directly be closed.
Hopefully not 🙏🏻

So I'm with my project where I would like to enforce the classes to be `readonly` in a directory
And since I know using the reflection you directly have access to this value, so I figured it would not be that a big of a deal 🧠

I figured it should look like something like that:

```php
$rules[] = Rule::allClasses()
        ->that(new ResideInOneOfTheseNamespaces('App\Domain\Event'))
        ->should(new IsReadonly())
        ->because('All event have to be readonly.')
    ;
```

I guess I should wait for your input before continuing in this way ?

Do you think this would be helpful for the community ?

Thank you for your time in any case and have a nice day !


fain182:
Thank you for your contribution @helyakin  !
I think this looks very good :slightly_smiling_face: 
If you could add the negative rule `IsNotReadOnly`, and a couple of lines in the README to show an example (like the other rules) and then we are ready for merging it!

helyakin:
Thanks for your quick review @fain182 

I'm super thrilled that you find this contribution interesting and worth adding 🙏🏻

I've just added the requested changes.

fain182:
great work, thanks!

Issue #338, opened 2023-01-14
How to exclude PHP Core classes? 


ghost:
### Support Question
I want to protect my domain from classes outside namespace but not the PHP core classes, e.g. \DateTimeImmutable etc.
How to I do that globally? I can exclude them for example passing then to NotHaveDependencyOutsideNamespace as second argument, but that's to much overhead.

veylor-fp:
Same with vendor code, 
for now we're using a whitelist which will become unmanageable over time

mxr576:
PHP Core classes would be also an interesting question to me, like `\DateTimeImmutable` or `\RuntimeException`. Should they be excluded by default?

PR #423, opened 2024-03-20
Add exclude root

notcgi:
https://github.com/phparkitect/arkitect/issues/338

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ $rules[] = Rule::allClasses()
 ```php
 $rules[] = Rule::allClasses()
     ->that(new ResideInOneOfTheseNamespaces('App\Domain'))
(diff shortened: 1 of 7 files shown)


AlessandroMinoccheri:
Hi @notcgi can you please fix the CI?

notcgi:
> Hi @notcgi can you please fix the CI?

done

fain182:
Exclude root namsepace is not the same thing as excluding the PHP built-in classes... but anyway is a solution we can evaluate.

AlessandroMinoccheri:
@notcgi when do you think is useful to exclude root classes?

notcgi:
> @notcgi when do you think is useful to exclude root classes?

Domain and Application layers can have dependencies like Datetime(-Immutable), Exception, Throwable, RuntimeException, UnitEnum etc

AlessandroMinoccheri:
Thank you @notcgi

PR #425, opened 2024-03-30
Fixed "IsAbstract" for interfaces, traits, enums and final classes

sebastianstucke87:
Fixed `\Arkitect\Expression\ForClasses\IsAbstract` because:

- interfaces can not be abstract
- traits can not be abstract
- enums can not be abstract
- final classes can not be abstract


AlessandroMinoccheri:
Hi @sebastianstucke87 can you please fix red checks?

AlessandroMinoccheri (approved):
LGTM 👍

courtney-miles:
Note that with this change, `IsAbstract` will now identify `interface`, `trait`, `enum` and `final` all as abstract classes.

> interfaces can not be abstract

I don't believe a rule is needed for this, because this is something PHP already disallows.

At best, I believe the rule could become...

```diff
- if ($theClass->isAbstract() || $theClass->isInterface() || $theClass->isTrait() || $theClass->isEnum()
-            || $theClass->isFinal()) {
+ if ($theClass->isAbstract() && !$theClass->isInterface() && !$theClass->isTrait() && !$theClass->isEnum()
+            && !$theClass->isFinal()) {
```
... but it should be redundant because if `::isAbstract()` is true then all other states must be `false`.

PR #430, opened 2024-05-01
Fix handling unordered baseline violations with ignore line numbers option

### LINKED ISSUE #428: Bug: Baseline with ignorelines not fully working
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Library Version | 0.3.29
| PHP version     | 8.3

#### Current behavior

Let's assume we have the following 

```
{
        {
            "fqcn": "App\Application\Foo",
            "line": 1,
            "error": "depends on App\\Ui\\Bar, but should not depend on these namespaces: App\\Ui\\*, App\\Tests\\* because of component architecture"
        },
        {
            "fqcn": "App\Application\Foo",
            "line": 3,
            "error": "depends on App\\Ui\\Baz, but should not depend on these namespaces: App\\Ui\\*, App\\Tests\\* because of component architecture"
        },
        {
            "fqcn": "App\Application\Foo",
            "line": 2,
            "error": "depends on App\\Ui\\Bar, but should not depend on these namespaces: App\\Ui\\*, App\\Tests\\* because of component architecture"
        },
}
```

What happens is that when doing the `phparkitect check --ignore-baseline-linenumbers`, the function used to remove in the `udiff` is check for the error and fqn first and then spaceship the sort -- and that messes up the cursor in one of the arrays and not fully cleaning up post doing it. In the case above we could be NOT removing the line 3 (BAZ) because it moved the cursor past that error when looking for BAR

#### How to reproduce

This is a little tricky, you basically need to create a baseline where the initial sorting using lines differ from the sort when ignoring lines. I can't share the current baseline where this happens because is from my company and I'm not allowed to do that.

#### Expected behavior

Everything that is on my baseline should be removed. Doing `phparkitect generate-baseline` and post `phparkictect check --ignore-baseline-numbers` should ALWAYS be successful. Because the initial sorting is taking in consideration lines, and later during comparing is not fully trusting that sort, it's breaking this removal.

marmichalski:
Handles unordered baseline violations when ignoring baseline numbers.
This resolves #428.

### DIFF
--- a/src/Rules/Violations.php
+++ b/src/Rules/Violations.php
@@ -116,7 +116,7 @@ class Violations implements \IteratorAggregate, \Countable, \JsonSerializable
             $this->violations = array_values(array_udiff(
                 $this->violations,
                 $violations->violations,
-                [__CLASS__, 'compareViolations'],
+                [__CLASS__, 'compareViolations']
             ));
 
             return;
--- a/src/Rules/Violations.php
+++ b/src/Rules/Violations.php
@@ -112,13 +112,30 @@ class Violations implements \IteratorAggregate, \Countable, \JsonSerializable
      */
     public function remove(self $violations, bool $ignoreBaselineLinenumbers = false): void
     {
-        $comparisonFunction = [__CLASS__, $ignoreBaselineLinenumbers ? 'compareViolationsIgnoreLineNumber' : 'compareViolations'];
+        if (!$ignoreBaselineLinenumbers) {
+            $this->violations = array_values(array_udiff(
+                $this->violations,
+                $violations->violations,
+                [__CLASS__, 'compareViolations'],
+            ));
 
-        $this->violations = array_values(array_udiff(
-            $this->violations,
-            $violations->violations,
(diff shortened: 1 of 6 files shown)


AlessandroMinoccheri:
Hi @marmichalski, thank you for this PR.
Can you please fix the Coding Standard errors?

AlessandroMinoccheri:
@marmichalski  Can you please add in the README file this new input `--ignore-baseline-linenumbers`? 

Thank you :)

marmichalski:
> @marmichalski Can you please add in the README file this new input `--ignore-baseline-linenumbers`?
> 
> Thank you :)

This one https://github.com/phparkitect/arkitect/blob/81711757459e2130a587e8eaca7da8b8e606ec37/README.md#line-numbers-in-baseline ?

AlessandroMinoccheri:
Oh right @marmichalski, forget about my question.
In my opinion the PR LGTM!

Thanks for your contribution!

AlessandroMinoccheri (approved):

PR #441, opened 2024-11-26
support php 8.4

nikow13:
updates the codebase for PHP 8.4 compatibility and upgrades Psalm to version 6.0. Some errors were introduced with the Psalm update, so I’ve added a baseline for now to prevent build issues


AlessandroMinoccheri:
Hello @nikow13!
Thanks for your contribution, we need to wait psalm I think to see if all tests will be green

damien-louis:
Related psalm issue and MR about PHP 8.4 support: 
https://github.com/vimeo/psalm/issues/11107
https://github.com/vimeo/psalm/pull/10928

ostrolucky:
Psalm 6 has been released which supports php 8.4

akondas:
ready to go?

ostrolucky:
looks like php version used for running scrutinizer needs to be bumped

nikow13:
> looks like php version used for running scrutinizer needs to be bumped

Yes, it seems Scrutinizer is managed outside the code, so I can’t update it

damien-louis:
Poke @AlessandroMinoccheri

AlessandroMinoccheri:
We are verifying it :)

Dockerfile
```diff
@@ -1,8 +1,8 @@
-ARG PHP_VERSION=7.1
+ARG PHP_VERSION=8.4
```
AlessandroMinoccheri, line 1: Why the version 8.4 and not the minor that we can support?

nikow13: for local testing with the last version

.github/workflows/build.yml
```diff
@@ -9,11 +9,11 @@ on:
 
 jobs:
   build:
-    runs-on: "ubuntu-22.04"
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        php-versions: [ '7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3']
+        php-versions: [ '8.2', '8.3', '8.4']
```
micheleorselli, line 16: we'd like to keep compatibility from php 8.0

.github/workflows/build.yml
```diff
@@ -52,22 +52,22 @@ jobs:
...
     needs: build
     steps:
       - uses: actions/checkout@v3
 
       - name: Install PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 7.1
-          tools: composer:v2.2
+          php-version: 8.4
```
micheleorselli, line 69: shouldn't we use the smaller supported version (8.0)?

.github/workflows/build.yml
```diff
@@ -52,22 +52,22 @@ jobs:
...
     - name: Upload coverage to Codecov
-      if: ${{ matrix.php-versions  == '7.1' }}
+      if: ${{ matrix.php-versions  == '8.4' }}
       uses: codecov/codecov-action@v1
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
 
   phar:
-    runs-on: "ubuntu-20.04"
+    runs-on: ubuntu-latest
```
micheleorselli, line 61: would that mean when ubuntu gets upgraded we automatically get the latest version? If that the case an approach where we explicitly set the version is preferred imo

.github/workflows/build.yml
```diff
@@ -84,19 +84,18 @@ jobs:
...
 
   smoke-test-phar:
     needs: phar
-    runs-on: "ubuntu-20.04"
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-versions: [ '7.1', '7.2', '7.3', '7.4', '8.0', '8.1' ]
-        coverage-driver: [ 'pcov' ]
+        php-versions: [ '8.2', '8.3', '8.4' ]
```
micheleorselli, line 98: same as above

Dockerfile
micheleorselli: that should be the smaller supported version maybe? Like 8.0 🤔

tests/Unit/Expressions/ForClasses/ResideInOneOfTheseNamespacesTest.php
```diff
@@ -35,7 +35,7 @@ public function shouldMatchNamespacesProvider(): array
      * @param mixed $actualFQCN
      * @param mixed $explanation
      */
-    public function test_it_should_match_namespace_and_descendants($expectedNamespace, $actualFQCN, $explanation): void
+    public function test_it_should_match_namespace_and_descendants(string $expectedNamespace, string $actualFQCN, string $explanation): void
```
micheleorselli, line 38: `@param` in phpdoc does not match with the method signature, would it be possibile to remove it?

src/CLI/TargetPhpVersion.php
```diff
@@ -17,10 +17,10 @@ class TargetPhpVersion
         '8.1',
         '8.2',
         '8.3',
+        '8.4',
```
micheleorselli, line 20: iirc here we are defining which version arkitect is capable of parsing. Should we keep versions older than 8.0? if yes, should then keep testing them (at least smoke test the phar) to be sure we are actually able to do so?

src/CLI/Progress/ProgressBarProgress.php
```diff
@@ -15,11 +15,9 @@
  */
 class ProgressBarProgress implements Progress
 {
-    /** @var OutputInterface */
-    private $output;
+    private \Symfony\Component\Console\Output\NullOutput|OutputInterface $output;
```
micheleorselli, line 18: isn't NullOutput implementing OutputInterface? If yes, that could be just 

`private OutputInterface $output;`

micheleorselli (changes requested):
Talking a bit with @AlessandroMinoccheri we agreed php 7 support can be dropped, but we would like to keep support for php 8.0 on

nikow13:
> Talking a bit with @AlessandroMinoccheri we agreed php 7 support can be dropped, but we would like to keep support for php 8.0 on

Hi,
There are a lot of inconsistencies between different PHP versions. I couldn't get it to work with 8.0 and 8.1, but feel free to give it a try if you have any ideas !

micheleorselli:
hey there, that should be fixed by #447

PR #442, opened 2025-01-02
Add compatibility with phpstan/phpdoc-parser v2

EmilMassey:


micheleorselli:
The build pipeline is fixed so by rebasing to master you should be able to get all the checks passing. 
Apart from that, what are the advantages of using version 2 of the parser? Consider we decided to drop php 7 support so we could just support version 2 and that's it. cc @AlessandroMinoccheri

micheleorselli:
@EmilMassey I guess we can just drop the support to version 1, would you mind updating the PR? Thank you!

EmilMassey:
While I don't believe any new features in v2 are necessary for this library, updating the dependency allows arkitect to be installed in codebases that need v2 for various reasons. That's the main reason for my PR. I updated the PR and dropped support for v1.

micheleorselli:
thanks @EmilMassey!

PR #443, opened 2025-01-03
Fix code style

EmilMassey:
Currently CI checks fail due to code style issues being reported by the latest version of CS fixer. This PR fixes these issues and also changes configuration for `trailing_comma_in_multiline` fixer not to add comma after parameters, because it's not allowed in PHP 7.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,6 +43,7 @@ jobs:
       run: composer install --prefer-dist
 
     - name: Coding Standard Checks
+      if: ${{ matrix.php-versions >= '8.0' }}
       run: PHP_CS_FIXER_IGNORE_ENV=1 ./bin/php-cs-fixer fix --dry-run -v
 
     - name: Static Analysis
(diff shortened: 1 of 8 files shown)


AlessandroMinoccheri:
Hi @EmilMassey  is it possible to fix pipelines?

EmilMassey:
Hi @AlessandroMinoccheri Does it make sense to run coding standard checks for PHP < 8.0? I added condition to skip this step in the pipelines, because the problem is that depending on PHP version, another CS Fixer version is installed and the configurations for both of them differ. Alternatively we could define different configuration options depending on the PHP version.

micheleorselli:
@EmilMassey we pinned phpcsfixer to a specific version so we should not get randomic results based on phpcsfixer and php version combinations. That said, I would like to bump phpcsfixer to the latest version and update the codebase to add return types. I would say this PR can be closed, what do you think?

EmilMassey:
Yes, I created this PR only to fix the failing CI checks, but bumping the CS Fixer and updating the codebase is definitely the way to go here. Thank you.

PR #450, opened 2025-02-11
Removes psalm as dev dependency and use it as a phar

micheleorselli:


szepeviktor:
@micheleorselli You can use both shivammathur/setup-php and PHIVE to get those tools as PHAR files.
No need to store (and update?) them in the repository.

PR #451, opened 2025-02-12
Update dependencies and improve type annotations

raffaelecarelle:
Added support for newer versions of `nikic/php-parser`, `vimeo/psalm`, and `phpunit/phpunit` in `composer.json`. Enhanced type annotations across multiple classes with `@implements` for iterator declarations and more explicit docblock types. Also replaced loose comparisons with strict ones for better code reliability.


micheleorselli:
Hey there, this should have been fixed by #447

micheleorselli:
would it be ok if I close this PR?

PR #454, opened 2025-02-26
Adds appliesTo to verify if a given rule should be applied

micheleorselli:
This is a tentative fix for https://github.com/phparkitect/arkitect/issues/434. It introduces a new method, appliesTo, which is used to decide if a given rules should be applied to the class under examination, eg:
- final classes, enums, interfaces, trait, cannot be abstract so we exclude them when we check if a given class is abstract or not, because they cannot have that property by definition. 
 

Caveat: with this fix a rule defined in that way

```php
$rule = Rule::allClasses()
            ->that(new ResideInOneOfTheseNamespaces('App\Abstract'))
            ->should(new IsAbstract())
            ->because('we want to prefix abstract classes');
```

would miss any final class present in the namesapce `App\Abstract`. Is that what we want?


fain182:
> would miss any final class present in the namespace App\Abstract. Is that what we want?

I don't think this would be the expected behaviour from our users....

src/Analyzer/FileVisitor.php
```diff
@@ -224,15 +224,6 @@ public function enterNode(Node $node): void
...
         }
-
-        if ($node instanceof Node\Attribute) {
-            $nodeName = $node->name;
-
-            if ($nodeName instanceof Node\Name\FullyQualified) {
-                $this->classDescriptionBuilder
-                    ->addDependency(new ClassDependency(implode('\\', $nodeName->getParts()), $node->getLine()));
-            }
-        }
```
micheleorselli, line 235: this snippet was adding the dependency derived by attributes twice

src/Expression/ForClasses/IsNotFinal.php
```diff
@@ -18,6 +18,11 @@ public function describe(ClassDescription $theClass, string $because): Descripti
         return new Description("{$theClass->getName()} should not be final", $because);
     }
 
+    public function appliesTo(ClassDescription $theClass): bool
+    {
+        return !($theClass->isInterface() || $theClass->isTrait() || $theClass->isEnum() || $theClass->isFinal());
```
fain182, line 23: this `isFinal` should not be there?

Issue #393, opened 2023-07-19
new Implement(\Throwable:;class) doesn't work as expected and has false positives


Et3rnel:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | yes?
| Library Version | 0.3.25 
| PHP version     | 8.2

#### Summary
I try to check if all my exceptions are suffixed by Exception keyword

#### Current behavior

All interfaces and traits seems to be flagged as violation

#### How to reproduce

Create this rule:
```php
$rules[] = Rule::allClasses()
    ->that(new Implement(\Throwable::class))
    // also doesn't work with ->that(new Implement('Throwable'))
    // also doesn't work with ->that(new Extend(\Exception::class))
    ->should(new HaveNameMatching('*Exception'))
    ->because('reasons');
```

#### Expected behavior
- The exceptions not suffixing Exception should be flagged as violations (that's not the case)
- All interfaces and traits should not be flagged as violations (because they don't event extend Throwable in my case)

micheleorselli:
Hey @Et3rnel, this should be fixed by #456 if you want to test it

PR #456, opened 2025-03-07
Fixes implements and extends rules behaviour

micheleorselli:
This should provide a fix for #393


AlessandroMinoccheri (commented):
Do we need to specify something more in the documentation @micheleorselli ?

micheleorselli:
> Do we need to specify something more in the documentation @micheleorselli ?

uhm, not really sure, we can probably specify the difference between `implements` and `extends` maybe?

PR #459, opened 2025-03-10
Refactor tests to use ClassDescriptionBuilder instead of new ClassDescription 

micheleorselli:
Using the builder is more coincise and readable and allows to evolve ClassDescription constructor without changing tests


fain182:
Great PR! The code is much better this way! 👏

PR #462, opened 2025-03-14
Remove version from composer.json

szepeviktor:
> Specifying the version yourself will most likely end up creating problems

https://getcomposer.org/doc/04-schema.md#version


szepeviktor:
for CLI\Version: https://getcomposer.org/doc/07-runtime.md#installed-versions

AlessandroMinoccheri (changes requested):
Thanks @szepeviktor for this PR, but if you would like to remove the version from the composer.json file, we should provide a way to print the version in a testable way.
And as you can see, VersionTest is broken.

szepeviktor:
How about adding version as constant here?
https://github.com/phparkitect/arkitect/blob/92bfae019e0e53df19c40fd0e2c08a610ea17074/src/CLI/Version.php#L7

fain182:
I have read the links above, but if the risk would be that of " Specifying the version yourself will most likely end up creating problems at some point due to human error."  then I don't see any advantage in writing the version in one place or another.

If we find a way to avoid writing the version in the repository, so we don't have to update it manually with every release, then that’s a different matter.

szepeviktor:
While testing it is not possible to get the *future* version number. (only existing tags can be queried)

While running:
```php
Composer\InstalledVersions::getVersion('phparkitect/arkitect');
```

micheleorselli:
> While testing it is not possible to get the _future_ version number. (only existing git tags can be queried)
> 
> While running:
> 
> ```
> Composer\InstalledVersions::getVersion('phparkitect/arkitect');
> ```

That should do the trick, we could use it in the Version class instead of parsing the composer.json., defaultingo to UNKNOWN if the class InstalledVersions does not exists.  BTW I guess the right package name is 

```
Composer\InstalledVersions::getVersion('phparkitect/phparkitect');
```

szepeviktor:
Google Gemini talks about these: https://g.co/gemini/share/dc76096d2ddc

szepeviktor:
Here it is 🎈

https://github.com/composer/composer/blob/71e8395ebdb68d9179fae53e7fa3a303d6e438f3/src/Composer/InstalledVersions.php#L171

szepeviktor:
It must be more complicated than I thought ...
```diff
--- Expected
+++ Actual
@@ @@
-'0.5.2'
+'dev-f3390ad0a999714bcc718754ad301fd8d8819716'
```

https://github.com/phparkitect/arkitect/blob/3acff7cab83d5e0fba24eecdc7e1a80909c69dea/tests/Unit/CLI/VersionTest.php#L14

micheleorselli:
> It must be more complicated than I thought ...
> 
> ```diff
> --- Expected
> +++ Actual
> @@ @@
> -'0.5.2'
> +'dev-f3390ad0a999714bcc718754ad301fd8d8819716'
> ```
> 
> https://github.com/phparkitect/arkitect/blob/3acff7cab83d5e0fba24eecdc7e1a80909c69dea/tests/Unit/CLI/VersionTest.php#L14

Considering that with this change everything is managed via composer, does this test still provide value? I'm for removing it

szepeviktor:
Done 🍏

szepeviktor:
Glad to contribute.

PR #463, opened 2025-03-14
Allow using phpdoc-parser v1.2

micheleorselli:
Version 1.2 is still used (for instance in the symfony demo project, so I am re-adding it


AlessandroMinoccheri (approved):
LGTM 👍

Issue #458, opened 2025-03-10
Track extends for Interfaces and Implements for Enums


rsavuliak-macpaw:
### Feature Request

Right now rules like `Extends`/`NotExtends` cannot be applied to Interfaces, because `FileVisitor` does not parse `extends` section from them. Same is about Enums and Implements section. 

```
if ($node instanceof Node\Stmt\Interface_) {
            if (null === $node->namespacedName) {
                return;
            }

            $this->classDescriptionBuilder->setClassName($node->namespacedName->toCodeString());
            $this->classDescriptionBuilder->setInterface(true);

            foreach ($node->attrGroups as $attributeGroup) {
                foreach ($attributeGroup->attrs as $attribute) {
                    $this->classDescriptionBuilder
                        ->addAttribute($attribute->name->toString(), $attribute->getLine());
                }
            }
        }
```

...however it has such info:
```
class Interface_ extends ClassLike {
    /** @var Node\Name[] Extended interfaces */
    public array $extends;
```

#### Summary

Pls add possibility to apply such combinations as Interfaces + Extends and Enums + Implements.

micheleorselli:
Thank you for reporting this, we'll work on it

AlessandroMinoccheri:
@rsavuliak-macpaw we merged the feature and release the new version, can you give us a feedback, please?

rsavuliak-macpaw:
now it works and detects interface extends 👍  Thanks a lot!

PR #464, opened 2025-03-14
Collect extends also for interfaces

micheleorselli:
This should implement #458. Since interfaces can extend multiple interfaces 🤯 the extend property is now an array

### DIFF
--- a/src/Analyzer/ClassDescriptionBuilder.php
+++ b/src/Analyzer/ClassDescriptionBuilder.php
@@ -8,47 +8,40 @@ use Webmozart\Assert\Assert;
 class ClassDescriptionBuilder
 {
     /** @var list<ClassDependency> */
-    private $classDependencies = [];
+    private array $classDependencies = [];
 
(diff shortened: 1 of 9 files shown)


src/Analyzer/ClassDescriptionBuilder.php
```diff
@@ -8,47 +8,40 @@
...
+    private ?FullyQualifiedClassName $FQCN = null;
 
     /** @var list<FullyQualifiedClassName> */
-    private $interfaces = [];
+    private array $interfaces = [];
 
-    /** @var ?FullyQualifiedClassName */
-    private $extend;
+    /** @var list<FullyQualifiedClassName> */
+    private array $extend = [];
```
AlessandroMinoccheri, line 19: Is it better to use the plural "extends" like for other properties and in other files?

micheleorselli: Yes, it makes sense 👍🏻

fain182 (approved):

Issue #460, opened 2025-03-11
Request: Sign releases with GPG


TomBrouws:
We're using PHPArkitect in our company, and since your releases are not signed, we're currently manually verifying the PHAR using a SHA256 hash.

Would you be open to add a GPG signature to the current/future releases? This will allow better and easier security, e.g. through [PHIVE](https://phar.io/). It's quite straightforward to do: https://phar.io/howto/sign-and-upload-to-github.html

micheleorselli:
Good idea, we use [Box](https://github.com/box-project/box) to package the phar and it seems kinda [straightforward to set up](https://github.com/box-project/box/blob/main/doc/phar-signing.md#how-to-properly-sign-your-phar)

micheleorselli:
hey there,
now releases starting from version [0.5.3](https://github.com/phparkitect/arkitect/releases/tag/0.5.3) are signed. Let us know if everything works as expected

TomBrouws:
Works like a charm, thanks @micheleorselli !

PR #465, opened 2025-03-18
Adds steps to sign phar when releasinng it

micheleorselli:
This address what is reported on #460

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,7 +84,7 @@ jobs:
 
       - name: "Rename phar to avoid conflicts"
         run: mv ./phparkitect.phar phparkitect-${{ github.sha }}.phar
-
+      
       - name: "Upload phar file artifact"
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         php-version: ${{ matrix.php-versions }}
         tools: composer:v2.2
       env:
-        GITHUB_TOKEN: ${{ secrets.TOKEN_GITHUB }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
micheleorselli, line 28: not strictly related to the scope of this PR but reading this [doc](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication) it seems the default token is enough. I haven't removed the `TOKEN_GITHUB` for now, let's see if we face problems

AlessandroMinoccheri (approved):
LGTM 👍

PR #467, opened 2025-03-20
Specify a custom temporary directory when creating the phar

micheleorselli:
This allows to change the default build dir when running `make phar `, like this

```
make phar TMP_DIR=~/workspace/tmp
```


AlessandroMinoccheri (approved):
LGTM 👍

PR #469, opened 2025-03-23
fix README.md

raffaelecarelle:
with string 'AsController' does not work


micheleorselli:
for the sake of consistency with the rest of the documentation I would keep the parameters as a string, maybe we should specify the fqcn?

AlessandroMinoccheri (approved):
LGTM 👍

PR #470, opened 2025-03-23
Support multiple output formats for violation reporting

raffaelecarelle:
Added support for violation reporting in both text and JSON formats by introducing a `Printer` interface and implementing `TextPrinter` and `JsonPrinter` classes. Refactored relevant methods to utilize the new `Printer` system and updated CLI options to allow format selection. Included comprehensive unit tests for the new functionality.

### DIFF
--- a/src/CLI/Command/Check.php
+++ b/src/CLI/Command/Check.php
@@ -26,6 +26,7 @@ class Check extends Command
     private const USE_BASELINE_PARAM = 'use-baseline';
     private const SKIP_BASELINE_PARAM = 'skip-baseline';
     private const IGNORE_BASELINE_LINENUMBERS_PARAM = 'ignore-baseline-linenumbers';
+    private const FORMAT_PARAM = 'format';
 
     private const GENERATE_BASELINE_PARAM = 'generate-baseline';
     private const DEFAULT_RULES_FILENAME = 'phparkitect.php';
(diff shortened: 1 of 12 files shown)


micheleorselli:
It looks good to me, just out of curiosity: is there a particular use case for the json output?

raffaelecarelle:
Hi @micheleorselli, for me the  scope of the feature is to use with gitlab ci and code quality format supported is json.

https://docs.gitlab.com/ci/testing/code_quality/#import-code-quality-results-from-a-cicd-job

thank you

raffaelecarelle:
> Hey @raffaelecarelle, before merging would you mind:
> 
>     1. moving the `Printer` directory under the `CLI` one?
> 
>     2. add an e2e test [here](https://github.com/phparkitect/arkitect/blob/main/tests/E2E/Cli/CheckCommandTest.php) to check the new option `--format=json `?
> 
> 
> Thanks 🙇

Hi, I realized that in order to have a valid json, it is necessary to suppress all irrelevant output. So I added another optione --only-errors to do this! 

As you suggested then, I added an e2e and moved the printer under the CLI folder.

Let me know.

micheleorselli:
my 2 cent: it feels a bit strange having to specify two options in order to get the right output. it would be nice that if I specify `--format=json` then I only get json as output, no additional option needed
@AlessandroMinoccheri @fain182 wdyt?

raffaelecarelle:
I too initially thought this but preferred to add an option for clarity and greater flexibility. Anyway you tell me, no problem with unifying.

raffaelecarelle:
ok guys, I remove the `--only-errors` option.

As you said, this make easier option json to use.

PR #471, opened 2025-03-24
Adds a test case in PatternStringTest

micheleorselli:
This adds a test case in the PatternStringTest, just to ensure the regression mentioned here #391 is fixed


AlessandroMinoccheri (approved):
LGTM 👍

PR #476, opened 2025-04-01
Removes $stopOnFailure params from Violations class

micheleorselli:
Moved the logic which manage the param `$stopOnFailure` in the `Runner` class


fain182 (approved):

AlessandroMinoccheri (approved):
LGTM 👍

PR #474, opened 2025-03-31
feat: Add GitLab output format support to CLI.

raffaelecarelle:
Introduced the "gitlab" format for CLI output, alongside existing "text" and "json" formats. This includes a new `GitlabPrinter` implementation, updates to the `PrinterFactory` and output option handling, as well as corresponding unit and E2E tests. The change enhances integration options, particularly for GitLab CI.

gitlab code quality format must have particular json format

see gitlab docs: https://docs.gitlab.com/ci/testing/code_quality/#code-quality-report-format

this PR depends on #475

PR #477, opened 2025-04-01
Add relative file path to classdescription

micheleorselli:
This depends on #476. Here a `$filePath` attribute is added to `ClassDescription` and is populated during the parsing process. That should enable adding the information relative to the file path when building a violation (needed in #474)

### DIFF
--- a/src/Analyzer/ClassDescription.php
+++ b/src/Analyzer/ClassDescription.php
@@ -74,11 +74,11 @@ class ClassDescription
         $this->enum = $enum;
     }
 
-    public static function getBuilder(string $FQCN, string $relativeFilePath): ClassDescriptionBuilder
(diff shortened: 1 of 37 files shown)


src/Analyzer/ClassDescription.php
```diff
@@ -70,10 +74,11 @@ public function __construct(
         $this->enum = $enum;
     }
 
-    public static function getBuilder(string $FQCN): ClassDescriptionBuilder
+    public static function getBuilder(string $FQCN, string $relativeFilePath): ClassDescriptionBuilder
```
AlessandroMinoccheri, line 77: Why everywhere we have filePath and here relativeFilePath?

AlessandroMinoccheri (approved):
LGTM 👍

PR #478, opened 2025-04-03
Violation can now contain a reference to the file path

micheleorselli:
This is the last step to unblock https://github.com/phparkitect/arkitect/pull/474. 
Now all rules populate the field `$filePath`, which can be used when generating the output.

### DIFF
--- a/src/Rules/Violation.php
+++ b/src/Rules/Violation.php
@@ -22,9 +22,9 @@ class Violation implements \JsonSerializable
         $this->filePath = $filePath;
     }
 
-    public static function create(string $fqcn, ViolationMessage $error): self
+    public static function create(string $fqcn, ViolationMessage $error, string $filePath): self
     {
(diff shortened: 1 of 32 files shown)


src/Rules/Violation.php
```diff
@@ -10,23 +10,26 @@ class Violation implements \JsonSerializable
 
     private ?int $line;
 
+    private ?string $filePath;
+
     private string $error;
 
-    public function __construct(string $fqcn, string $error, ?int $line = null)
+    public function __construct(string $fqcn, string $error, ?int $line = null, ?string $filePath = null)
```
fain182, line 17: maybe we could make this construct private?

fain182 (approved):

src/Rules/Violation.php
micheleorselli: doable, will tackle in a separate PR 🙇

PR #484, opened 2025-04-12
Print violations to stdout, everything else to stderr

micheleorselli:
Here the output is divided between STDOUT and STDERR, in particular:
- the list of violations is printed on stdout
- everything else on stderr

This has several benefits
- we can remove the `$onlyError` logic scattered in the code
- the output now can be piped to other commands or in a file and in case of json/gitlab format you end up having a valid json


AlessandroMinoccheri (approved):
LGTM 👍

PR #489, opened 2025-04-24
Check: adds an option to specifiy an autoloader file to be included

micheleorselli:
**Context:** as we are considering the possibility to use functions like `is_a` in the rules, we need a way to deal with the autoload of the classes phparkitect is analizyng

We have two scenarios based on how phparkitect is installed

**as a project dependency via composer:** this is the simplest case since, since phparkitect shares the autoload with the project it analyzes, no additional autoloader is needed ✅ 

**as a phar:** here phparkitect autoload is embedded in the phar itself, so we need a way to autoload the project's classes

In the second scenario, we should consider the phar could be in different locations: 
- global, like in `/usr/local/bin/phparkitect`
- in the project root
- in the project `bin` dir
- ...

so trying to guess where the project autoload could be can be challenging. Possible solutions to tackle that:
1. do nothing and keep things as they are 😢 .
2. improve the parser to recursively fetch ancestors for classes, interfaces, enum, traits.
3. allow to specify an extra autoloader. 

Solution no 2. would be the most consistent with the current approach as the ClassDescription object would contain all the info needed to apply a rule and no autoload will be triggered. This is also the most complex to implement properly and it may need rethinking the architecture which now assumes we parse a file -> create one or more ClassDescription objects -> apply rules on them. 

Solution no.3 is kinda common on other tools ([rector](https://getrector.com/documentation/static-reflection-and-autoload), [psalm](https://psalm.dev/docs/running_psalm/configuration/#autoloader), [phpstan](https://phpstan.org/user-guide/discovering-symbols#custom-autoloader)) so it should not be a big surprise for the users. It is also the easiest to implement. _Is there any chance of getting conflicts between phparkitect's classes and the project classes?_ No, because the tool we use to package the phar uses [php-scoper](https://github.com/humbug/php-scoper) to avoid this kind of problems

**Choosen solution:** while it could be nice to do spike in order to understand the effort of implementing 2. it properly here I went for solution no 3.: _adding the capability to specify the path of an extra autoloader file via command line option_. 

this should allow what is asked in #401

### DIFF
--- a/src/CLI/Command/Check.php
+++ b/src/CLI/Command/Check.php
@@ -141,10 +141,10 @@ class Check extends Command
                 ->skipBaseline($skipBaseline)
                 ->format($format);
 
-            $this->requireAutoload($config->getAutoloadFilePath(), $output);
-            $printer = $this->createPrinter($config->getFormat(), $output);
-            $progress = $this->createProgress($verbose, $output);
-            $baseline = $this->createBaseline($config->isSkipBaseline(), $config->getBaselineFilePath(), $output);
+            $this->requireAutoload($output, $config->getAutoloadFilePath());
+            $printer = $this->createPrinter($output, $config->getFormat());
+            $progress = $this->createProgress($output, $verbose);
+            $baseline = $this->createBaseline($output, $config->isSkipBaseline(), $config->getBaselineFilePath());
 
             $output->writeln("Config file '$rulesFilename' found\n");
 
@@ -189,7 +189,7 @@ class Check extends Command
     /**
      * @psalm-suppress UnresolvableInclude
      */
-    protected function requireAutoload(?string $filePath, OutputInterface $output): void
+    protected function requireAutoload(OutputInterface $output, ?string $filePath): void
     {
         if (null === $filePath) {
             return;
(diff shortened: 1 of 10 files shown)


fain182:
We need to add this option to the readme... maybe in the phar section?

src/CLI/Command/Check.php
```diff
@@ -95,6 +99,12 @@ protected function configure(): void
                 InputOption::VALUE_OPTIONAL,
                 'Output format: text (default), json, gitlab',
                 'text'
+            )
+            ->addOption(
+                self::AUTOLOAD_PARAM,
+                'a',
+                InputOption::VALUE_REQUIRED,
```
AlessandroMinoccheri, line 106: Why this param is required and not optional?

micheleorselli: That's because if you specify --autoload you need to provide a value for it

src/CLI/Command/Check.php
```diff
@@ -177,6 +186,45 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         }
     }
 
+    /**
+     * @psalm-suppress UnresolvableInclude
+     */
+    protected function requireAutoload(?string $filePath, OutputInterface $output): void
```
AlessandroMinoccheri, line 192: Why not invert params to have the nullable value at the end?

src/CLI/Command/Check.php
```diff
@@ -177,6 +186,45 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+    }
+
+    protected function createProgress(bool $verbose, OutputInterface $output): Progress
+    {
+        $output->writeln('Progress: '.($verbose ? 'debug' : 'bar'));
+
+        return $verbose ? new DebugProgress($output) : new ProgressBarProgress($output);
+    }
+
+    protected function createBaseline(bool $skipBaseline, ?string $baselineFilePath, OutputInterface $output): Baseline
```
AlessandroMinoccheri, line 219: same as above.

src/CLI/Command/Check.php
micheleorselli: Yup, I guess so

AlessandroMinoccheri (approved):
LGTM 👍

Issue #500, opened 2025-05-08
Add support to exclude sub-namespaces in NotDependsOnTheseNamespaces rule


pfazzi:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

I would like to exclude specific sub-namespaces from the `NotDependsOnTheseNamespaces` rule.

Right now, I use this rule like this:
```php
Rule::allClasses()
   ->that(new ResideInOneOfTheseNamespaces('App'))
   ->should(new NotDependsOnTheseNamespaces('App\Foo')
   ->because("...");
```

I need a way to skip some nested namespaces (e.g. `App\Foo\PublicApi` should be allowed as a dependency).

One option could be:
```php
Rule::allClasses()
   ->that(new ResideInOneOfTheseNamespaces('App'))
   ->should(new NotDependsOnTheseNamespaces('App\Foo', except: ['App\Foo\PublicApi'])
   ->because("...");
```

micheleorselli:
@pfazzi can we consider it closed?

Issue #480, opened 2025-04-07
How to ensure module independence with phparkitect in legacy systems?


tomaszhanc:
|    Q            |   A
|---------------- | ------
| Library version | 0.5.3.0
| PHP version     | 8.4.1

## Support Question

I'm new to phparkitect and would like to know the best way to configure it for my use case. I'm open to creating a PR if something should be added to the codebase 🙌.

### Big Ball of Mud into Modular Monolith

I have a legacy application that's essentially a big ball of mud. **Everything is coupled to everything else**. I want to start adding some separation between modules. To begin, I plan to ensure there's a single point of contact, like through a Module Facade. I tried to define the below rules:

**Rule 1: Other modules should access the **`Shop`** module only through **`ShopFacade`

```
$rules[] = Rule::allClasses()
    ->that(new ResideInOneOfTheseNamespaces(''))
    ->andThat(new NotResideInTheseNamespaces('App\Shop'))
    ->should(new NotDependsOnTheseNamespaces('App\Shop'))
    // how to exclude ShopFacade?
    ->because('Other modules should only access Shop through ShopFacade');
```

In an ideal situation, this would be enough, but I won't have all the modules defined initially. I will add them gradually, so I also need a rule to ensure that the extracted module doesn't have coupling with other modules that don't yet have rules in phparkitect.

**Rule 2: `Shop` module can only rely on its own or shared classes, and the ProductFacade.**

```
$rules[] = Rule::allClasses()
    ->that(new ResideInOneOfTheseNamespaces('App\Shop'))
    ->should(new NotHaveDependencyOutsideNamespace('App\Shop', ['App\Product\ProductFacade', 'App\Shared\*']))
    ->because('Modules should access other modules only through their public facades');
```

### Questions

- How can I exclude `ShopFacade` as a dependency in Rule 1? Should it stay in the `App\Shop` namespace, or should it be moved outside?
- Is there a better way to achieve my goal? I have over 20 modules in that repository, which could lead to a messy configuration. I considered creating a `builder` to avoid copy-pasting, but perhaps there's a more effective method in phparkitect.

AlessandroMinoccheri:
I agree with @micheleorselli.
Adding the possibility to exclude particular namespaces or classes can be the right way, and we did it already for other expressions like for: `NotHaveDependencyOutsideNamespace`.

An alternative way if you don't have too many namespaces is to use the expression: `DependsOnlyOnTheseNamespaces`

```
[...]

PR #501, opened 2025-05-08
Allow to exclude some namespaces in NotDependsOnTheseNamespaces and DependsOnlyOnTheseNamespaces

### LINKED ISSUE #481: Exclude checks in specific namespaces
### Feature Request

Add the possibility to exclude namespaces from expressions

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | it depends

#### Summary

After further discussion in this issue: https://github.com/phparkitect/arkitect/issues/480, it became clear that there's currently no standard or intuitive way to exclude specific namespaces in expressions like `NotDependsOnTheseNamespaces`.

This expression uses a variadic parameter in its constructor, which makes it difficult to introduce additional arguments—such as an "exclude" array to define namespaces that should be ignored during the check.

We currently see two potential paths forward:

1. Introduce a breaking change: Replace the variadic parameter with a standard array, allowing us to introduce a second parameter (e.g., for excluded namespaces). To maintain consistency, this change would need to be applied across all similar expressions, which could break existing implementations.
2. Introduce a new rule, such as `ExcludeTheseNamespacesFromChecks`, which would accept an array (or variadic arguments) of namespaces to be excluded. This would avoid breaking changes and keep existing rules intact.

Example:
```php
$rules[] = Rule::allClasses()
    ->that(new ResideInOneOfTheseNamespaces('')) // yes, some classes are not namespaced
    ->andThat(new NotResideInTheseNamespaces('App\Shop'))
    ->should(new NotDependsOnTheseNamespaces('App\Shop'))
    ->exclude(new ExcludeTheseNamespacesFromChecks('App\Shop\ShopFacade'))
    ->because('Other modules should only access Shop through ShopFacade');
```

Or 
```php
$rules[] = Rule::allClasses()
    ->that(new ResideInOneOfTheseNamespaces('')) // yes, some classes are not namespaced
    ->andThat(new NotResideInTheseNamespaces('App\Shop'))
    ->should(new NotDependsOnTheseNamespaces('App\Shop'))
    ->excludeTheseNamespacesFromChecks('App\Shop\ShopFacade')
    ->because('Other modules should only access Shop through ShopFacade');
```

What do you think @fain182 @micheleorselli @pfazzi ?

micheleorselli:
Here we are introducing a BC break to add this capability, This should fix #481 #480 #500

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ Currently, you can check if a class:
 ```php
 $rules[] = Rule::allClasses()
     ->that(new ResideInOneOfTheseNamespaces('App\Domain'))
-    ->should(new DependsOnlyOnTheseNamespaces(['App\Domain', 'Ramsey\Uuid'], ['App\Exluded']))
+    ->should(new DependsOnlyOnTheseNamespaces(['App\Domain', 'Ramsey\Uuid'], ['App\Excluded']))
     ->because('we want to protect our domain from external dependencies except for Ramsey\Uuid');
 ```
 
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ Currently, you can check if a class:
 ```php
 $rules[] = Rule::allClasses()
     ->that(new ResideInOneOfTheseNamespaces('App\Domain'))
-    ->should(new DependsOnlyOnTheseNamespaces('App\Domain', 'Ramsey\Uuid'))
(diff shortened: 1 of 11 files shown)


pfazzi (approved):
LGTM! Thanks! 🙏

AlessandroMinoccheri (approved):
LGTM 👍 
Only the little fix in the READMe to fix, please

Issue #497, opened 2025-05-07
Simplify parsing of docblock


micheleorselli:
In order to add dockblock parsing capabilities we are overriding a PhpParser class, namely `NameResolver`. This makes upgrading PhpParser more difficult as we need to repatch the NameResolver class every time. 

It could be worth exploring alternatives solution which do not require this hack, just to name a couple:
- create a smaller, simplified version of NameResolver targeting only docblocks
- extend NameResolver instead of patching it

PR #504, opened 2025-05-13
Use custom resolver for extracting type information from docblocks

micheleorselli:
This is a possibile approach to solve #497: the code needed to parse docblock types is now in a class indipendent from PhpParser's NameResolver. This would give us more flexibility and mantainability . 

Further evolution could be (to be tackled in subsequent PRs)
- extracting information for @throws tags and consider them as dependencies
- simplify the way we get information on types from the docblocks


AlessandroMinoccheri (approved):
Despite the error in the pipeline, for me it's fine.

LGTM 👍

PR #507, opened 2025-05-15
Collect types declared in @throws tags and consider them dependencies (1/2)

micheleorselli:
This first step is a refactoring to simplify `DocblockResolver` introducing some new support classes 

One thing I noticed during this refactoring is even in case of multiple types defined, like 

```php
/**
* @Assert\Blank
* @var Foo
*/ 
public $foo
``` 

we collect just one. Is that intended? do we want to address it in some way? We could support it using union types 
cc @fain182  @AlessandroMinoccheri?


AlessandroMinoccheri:
@micheleorselli for me, we should fix it in another PR, or at least try to understand deeper the context.

AlessandroMinoccheri (approved):
LGTM 👍

PR #508, opened 2025-05-20
Fix FileVisitor issues with wrong class detection

hgraca:
It seems the php parser is using phpdocs and deciding that
 something like
/**
 * @return array<int, string|int>
 * / 
 is a class with FQCN `My\Namespace\array<int, string|int>` which then ofc fails to create a FullyQualifiedClassName.

### DIFF
--- a/src/Analyzer/FileVisitor.php
+++ b/src/Analyzer/FileVisitor.php
@@ -247,6 +247,10 @@ class FileVisitor extends NodeVisitorAbstract
             return;
         }
 
+        if (FullyQualifiedClassName::isNotAValidFqcn($type->toString())) {
+            return;
+        }
+
         $this->classDescriptionBuilder
             ->addDependency(new ClassDependency($type->toString(), $node->getLine()));
     }
@@ -314,6 +318,10 @@ class FileVisitor extends NodeVisitorAbstract
             return;
         }
 
+        if (FullyQualifiedClassName::isNotAValidFqcn($returnType->toString())) {
+            return;
+        }
+
         $this->classDescriptionBuilder
             ->addDependency(new ClassDependency($returnType->toString(), $returnType->getLine()));
     }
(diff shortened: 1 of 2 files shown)


micheleorselli:
Hey @hgraca, thank you for reporting that 🙇. I guess this depends on the new `DocblockParser` and related classes. We'll try to look into it asap, in the meantime, if you want to give it a try the fix should happen in the `DocblockParser` and/or `DocblockTypesResolver` classes.

micheleorselli:
https://github.com/phparkitect/arkitect/pull/510 should fix the problem

AlessandroMinoccheri (approved):
LGTM

hgraca:
@micheleorselli thank you for fixing the bug, is this PR still worth merging or shall I close it?

micheleorselli:
Hey @hgraca thank you for testing it! Based on the error it seems this is related to something different rather than the `@return` tag. It seems the problem is caused by `@psalm-var`, right? Anyway we should exclude all the docblock tags we are not interested in parsing

hgraca:
@micheleorselli 

I finally had some time to look into what situations still give problems, I found only this:

https://github.com/ramsey/uuid/blob/4.x/src/Uuid.php#L232-L234

We updated our project from php8.2 to 8.4 this week, including some libraries ofc so I guess that's why this is not exactly the same issue as before, but at the root it seems to be the same, it looks like it interprets `@my-custom-doc-tag` as a class FQCN.

Unfortunately I dont know what shoulod be done to solve this on phparkitect side, except for what is in this PR already :(

let me know if I can be of any further assistance.

micheleorselli:
I was able to reproduce it, this https://github.com/phparkitect/arkitect/pull/517 should fix it

fain182:
@hgraca we can close this issue?

PR #512, opened 2025-05-21
FileVisitor improvements

micheleorselli:
some more refactorings before tackling the @throws tag dependency collection 

1. refactored to avoid nested conditionals
2. preferred checks like `$node->class instanceof Node\Name\FullyQualified` instead of `method_exists($node->class, 'toString')`. This allows to skip checks on parent:: self:: static:: and native types since a `Node\Name\FullyQualified` refers always to a class


AlessandroMinoccheri (approved):
LGTM

PR #513, opened 2025-05-22
Collect dependencies defined in @throws docblock tags

micheleorselli:


fain182:
fixes #514

fain182:
to complete this we should fix the row number, that now is the first line of the comment instead of the line of the throws.

Issue #303, opened 2022-10-23
Duplicated errors when `that()` is reused for multiple `should()`


Wirone:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no?
| Library Version | 0.2.32
| PHP version     | 7.4.32

#### Summary

Consider such configuration:

```php
return static function (Config $config): void {
    $srcClassSet = ClassSet::fromDir(__DIR__ . '/src');
    $rules = [];

    $rectors = Rule::allClasses()->that(new ResideInOneOfTheseNamespaces('Codito\Rector\Money\Rule'));
    $rules[] = $rectors->should(new HaveNameMatching('*Rector'))->because('this is Rector convention');
    $rules[] = $rectors
        ->should(new Extend(AbstractRector::class))
        ->because('we need satisfy Rector\'s contract for rules');

    $config->add($srcClassSet, ...$rules);
};
```

Since result of the `should()` is a `BecauseParser` it does not allow multiple expectations for the same selection, I wanted to reuse result of `that()` to make 2 separate rules. Unfortunately, it does not behave as I expected.

#### Current behavior

For class that does not satisfy both rules (empty `Codito\Rector\Money\Rule\Foo` class) I get:

```
Codito\Rector\Money\Rule\Foo has 4 violations
  should have a name that matches *Rector because this is Rector convention
  should extend Rector\Core\Rector\AbstractRector because this is Rector convention
  should have a name that matches *Rector because we need satisfy Rector's contract for rules
  should extend Rector\Core\Rector\AbstractRector because we need satisfy Rector's contract for rules
```

#### How to reproduce

- Fork [`codito/rector-money`](https://github.com/CoditoNet/rector-money/)
- Create empty class in `src/Rule` directory
- Run `make arkitect`

#### Expected behavior

I would expect only 2 errors for such scenario:

```
Codito\Rector\Money\Rule\Foo has 2 violations
  should have a name that matches *Rector because this is Rector convention
  should extend Rector\Core\Rector\AbstractRector because we need satisfy Rector's contract for rules
```

Ideally, it would be great to be able to define rule like this:

```php
return static function (Config $config): void {
    $srcClassSet = ClassSet::fromDir(__DIR__ . '/src');
    $rules = [];

    $rules[] = Rule::allClasses()
        ->that(new ResideInOneOfTheseNamespaces('Codito\Rector\Money\Rule'))
        ->should(new HaveNameMatching('*Rector'))
[...]

PR #543, opened 2025-12-14
Permits to reuse the same that() for multiple should()

fain182:
This fixes issue #303 where reusing the result of that() for multiple should() calls would cause duplicate violations due to shared mutable state in the RuleBuilder.

The fix makes should() and andThat() clone the RuleBuilder before adding constraints/specs, ensuring each rule branch gets an isolated copy of the builder state.

Added RuleBuilder::__clone() to deep clone Specs and Constraints objects.


AlessandroMinoccheri:
I would suggest adding few lines in the documentation.

PR #554, opened 2025-12-27
Add rule for exact matching of namespaces

### LINKED ISSUE #406: Support for checking the namespace exactly
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Currently (AFAIK) the only way to verify namespaces is `ResidesInTheseNamespaces` which is always recursive. There's use cases where we want to check the namespace **exactly**, for example `App\Entity`, not `App\Entity\Whatever`.

fain182:
This commit adds two new expression rules to support exact namespace matching, addressing the feature request in issue #406:

- ResideInOneOfTheseNamespacesExactly: Validates that classes reside in one of the specified namespaces exactly, without matching child namespaces
- NotResideInOneOfTheseNamespacesExactly: The inverse rule that validates classes do NOT reside in the specified namespaces exactly

The existing rules (ResideInOneOfTheseNamespaces and NotResideInTheseNamespaces) match recursively, including all child namespaces. The new rules provide an exact match option for stricter architectural constraints.

fixes #406

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,6 +400,36 @@ $rules[] = Rule::allClasses()
     ->because('we want to be sure that all events not reside in wrong layers');
 ```
 
+### Reside in a namespace exactly
+
+This rule checks that classes reside **exactly** in one of the specified namespaces, without matching child namespaces. Unlike `ResideInOneOfTheseNamespaces` which matches recursively, this rule only matches classes directly in the namespace.
+
+```php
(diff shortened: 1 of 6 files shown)


AlessandroMinoccheri (approved):
LGTM 👍

Issue #169, opened 2021-10-01
Inheritance issues with Implement and Extend


benr77:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Library Version | 0.2.0
| PHP version     | 8.0

#### Summary

Do the `Extend` and `Implement` classes handle multiple levels of inheritance?

E.g. I have this line, to pick up Symfony form types:

```
$formTypeClassNames = Rule::allClasses()
        ->that(new ResideInOneOfTheseNamespaces('App\UI\Form'))
        ->andThat(new Implement(FormTypeInterface::class))
        ->should(new HaveNameMatching('*Type'))
        ->because('we want uniform form type naming');
```

However, it's ignoring all these classes, because they _extend_ `AbstractType` which in turn _implements_ `FormTypeInterface`

#### Expected behavior

I would expect the check to consider interfaces implemented, or classes extended from, all the way up the inheritance tree.

Thanks

AlessandroMinoccheri:
Hi @benr77 you are right, I created a little test about the behavior and we need to check inside extend class other dependencies.

Broken test to validate the expected behavior:

```
public function test_it_should_parse_extends_class(): void
    {
        $code = <<< 'EOF'
<?php

namespace Root\Animals;

class Animal implements Foo
{
}

class Cat extends Animal
{

}
EOF;

        /** @var FileParser $fp */
        $fp = FileParserFactory::createFileParser();
        $fp->parse($code);

        $cd = $fp->getClassDescriptions()[1];

        $this->assertEquals(['Foo'], $cd->getInterfaces());
    }
```

lchrusciel:
Hey folks! 

thanks for your amazing work! Recently, we've introduced phparkitect in Sylius - https://github.com/Sylius/Sylius/pull/13426. It worked really nice. After few days, I've noticed, that we had one class out of namespace scope, so I fixed it and try to protect us in the future with creation of a new rule - https://github.com/Sylius/Sylius/pull/13495/commits/7772229adc4e5f181ca03748cb66288e8d66f428

What I would like to achieve, is to be sure that everything that implements `ApiPlatform\Core\Api\FilterInterface` (which is class from vendor) will be placed in `Sylius\Bundle\ApiBundle\Filter` namespace. The issue is, most of these classes do not implement this interface directly but through class inheritance. I was hoping that with phparkitect I would be able to enforce such rules even if they are based on vendor classes - evne tho, I can imagine that this increase complexity a lot.

[...]

PR #571, opened 2026-02-18
Check inheritance transitively in ForClasses expressions

fain182:
## Summary

- `Implement` / `NotImplement` now detect interfaces inherited from parent classes, not just those declared directly on the class under analysis.
- `Extend` / `NotExtend` now walk the full ancestor chain, catching indirect inheritance.
- `HaveTrait` / `NotHaveTrait` now collect traits used anywhere in the hierarchy, including parent classes.

Previously these rules only looked at what the static parser had extracted from the file, so a class that inherited an interface or trait from a parent would not be matched. Now they use `ReflectionClass` at evaluation time, which gives the complete picture.

As a consequence, classes must be autoloadable. If reflection fails, a `ReflectionException` is raised — no more silent skips that could hide misconfigured autoloading.

`composer dump-autoload` required after checkout

### DIFF
--- a/tests/Unit/Analyzer/FileParser/CanParseEnumsTest.php
+++ b/tests/Unit/Analyzer/FileParser/CanParseEnumsTest.php
@@ -32,6 +32,33 @@ class CanParseEnumsTest extends TestCase
     }
 
     /**
+     * @requires PHP 8.1
+     */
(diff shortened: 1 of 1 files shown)


fain182:
fixes #169

PR #573, opened 2026-02-19
Add a IsNotA expression as a symmetry to the existing IsA expression.

sukei:
I created a `IsInstanceOf` expression for my own project that is roughly the same implementation as the new `IsA` expression. But I also use a `IsNotInstanceOf` which have no equivalent yet. This PR add the `IsNotA` expression that is the exact inverse of the existing `IsA` implementation. I also rewrite the violations message to better align with the format of other expressions and that works better with the `because` suffix that may be append.


fain182:
Great, thanks!

PR #578, opened 2026-02-26
Update webmozart/assert dependency to version 2.0

chadyred:
Hello,

This pull request updates the dependency webmozart/assert to version 2, which is now officially available.

Please let me know if any additional adjustments are required.

Thanks 🙂


fain182:
Hi Florian, thanks for your PR!
I think we could support both versions, similar to how we support multiple PHPUnit versions.

PR #581, opened 2026-03-01
Make baseline resilient to rule configuration changes

### LINKED ISSUE #377: Baseline should still work when allowed namespaces are changed
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

The baseline stores the full error message, which in case of namespace rules contains the list of all allowed namespaces. When the configuration is changed, all baseline entries no longer match.

The part that explains what is allowed `...but should depend only on classes in one of these namespaces ...` should not be part of the baseline and not be required to match.

My idea would be to split the exception message into the actual `error` and a `hint` part. The hint should not be stored in the baseline and not be used when comparing with the baseline.

fain182:
## Summary
This change makes violation baseline matching more resilient to rule configuration changes by extracting and comparing only the stable violation-specific part of error messages, ignoring configuration-dependent details like allowed namespaces. Fixes #377 

## Key Changes
- Added `extractViolationKey()` method that separates the stable violation part from the rule description in error messages
  - For messages with format "$violation, but $ruleDescription": extracts only the "$violation" part
  - For self-explanatory messages without ", but ": returns the full message
- Updated `remove()` method to use `extractViolationKey()` when comparing violations against baseline
- Updated `compareViolations()` method to use the extracted violation key for consistent comparison
- Updated `compareViolationsIgnoreLineNumber()` method to use the extracted violation key

## Implementation Details
The violation message format distinguishes between:
1. **WithDescription format**: "$violation, but $ruleDescription" (e.g., "depends on App\Bar, but should depend only on classes in one of these namespaces: App\Domain")
2. **SelfExplanatory format**: "$ruleDescription" without ", but " (e.g., "should be final because we want immutability")

By extracting only the violation-specific part before the ", but " separator, baseline entries remain valid even when rule configuration changes (e.g., when allowed namespaces are updated). This prevents false positives when comparing violations against baselines.

## Tests Added
- `test_remove_violations_matches_when_rule_description_changes`: Verifies matching works when rule description changes
- `test_remove_violations_matches_when_rule_description_changes_ignore_linenumber`: Verifies matching with line number ignored
- `test_remove_violations_does_not_match_different_dependency`: Ensures different violations don't match
- `test_remove_violations_still_works_for_self_explanatory_messages`: Ensures self-explanatory messages still work correctly

https://claude.ai/code/session_01DPotRHDJzXnmPGPoCVp8Rn

### DIFF
--- a/src/Rules/Violations.php
+++ b/src/Rules/Violations.php
@@ -142,21 +142,6 @@ class Violations implements \IteratorAggregate, \Countable, \JsonSerializable
     }
 
     /**
-     * Comparison method that only checks the namespace and error but ignores the line number.
-     */
-    public static function compareViolationsIgnoreLineNumber(Violation $a, Violation $b): int
-    {
-        if (
-            $a->getFqcn() === $b->getFqcn()
-            && self::extractViolationKey($a->getError()) === self::extractViolationKey($b->getError())
-        ) {
-            return 0;
-        }
-
-        return self::compareViolations($a, $b);
-    }
-
-    /**
      * Extracts the stable violation-specific part from an error message.
      *
      * ViolationMessage produces two formats:


micheleorselli (approved):
looks good to me, maybe we should review how we deal with these errors as it is not really straightforward

PR #587, opened 2026-03-08
Support for xdebug in local development with docker

ArgusKay99:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | yes

#### Summary
Can we add xdebug to the docker-container? I created a multi-stage Dockerfile that can be built with and without xdebug.

For context: I am developing on linux with Phpstorm and use the the Docker-container that is created in the CONTRIBUTING.md file as remote PHP interpreter and when there is a xdebug installed, i can already debug unit-tests.

I'm open to suggestions to use xdebug in another way.

BC Break for local development:
docker image build will now need a target: `docker image build --target php_build -t phparkitect . `

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,10 +24,8 @@ ENTRYPOINT [ "phparkitect"]
 FROM php_build AS with_xdebug
 
 # compatibility chart: https://xdebug.org/docs/compat
-ARG XDEBUG_VERSION="xdebug-3.4.0"
+ARG XDEBUG_VERSION="xdebug-3.4.7"
 RUN apk add autoconf g++ linux-headers
 
-RUN pecl install ${XDEBUG_VERSION} \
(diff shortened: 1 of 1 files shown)


ArgusKay99:
Hey @fain182 , oh thanks for noticing. I fixed this.

Debugging the unit-tests works fine. This is because apparently my ide did set all the correct environment variables. So as long xdebug is installed and configured in php.ini it will work. 

image

Another Thing i noticed is that the tests can run and can be debugged, but will be shown as terminated. This seems to be the problem because the entrypoint is phparkitect (which i cannot change without destroying some workflows) so i'm unsure if it is smart to run xdebug in the same container? Out of curiosity how are you debugging php-arkitect? Maybe i'm just working "wrong".
image

PR #607, opened 2026-04-04
perf: cache ReflectionClass calls in isPhpCoreClass

### LINKED ISSUE #580: isPhpCoreClass() is causing significant performance overhead
## Description

The `isPhpCoreClass()` method in `FileVisitor` instantiates a `ReflectionClass` for every dependency
encountered during AST traversal. For classes that don't exist at analysis time, it relies on catching
`ReflectionException` as control flow.

This method was introduced after v0.8.0 and is the single largest contributor to the performance
regression observed between v0.8.0 and the current (unreleased) version.

## Impact

Profiling on arkitect's own codebase (105 files, ~17,700 AST nodes):

- `FileVisitor` accounts for **~29ms** of traversal time (34% of AST traversal)
- `isPhpCoreClass()` is estimated to cause **~32ms** of overhead (including exception handling cost)
- This represents approximately **60% of the total ~53ms regression** vs v0.8.0

The cost scales linearly with the number of dependencies in the analyzed codebase. On larger projects
the impact will be proportionally greater.

fain182:
Closes #580

## Problem

`isPhpCoreClass()` in `ClassDescriptionBuilder` calls `new ReflectionClass()` for every dependency of every analyzed class, with no caching. This was identified in #580 as the source of a measurable performance regression introduced after v0.8.0.

On arkitect's own 104-file codebase, `ReflectionClass` was being instantiated **849 times** — many of those for the same class names repeated across files, and many more for user-land classes that don't even exist in the current process (triggering exception handling as control flow).

## Changes

- **Cache**: results are stored in a static array keyed by FQCN. Subsequent calls for the same class name return immediately without any reflection.
- **Early exit**: if the class is not loaded in the current process (`class_exists` / `interface_exists`), it cannot be a PHP internal class — return `false` immediately, skipping `ReflectionClass` entirely.
- **Removed try/catch**: now that we guard with `class_exists` first, `ReflectionClass` will never throw for the classes that reach it, so exception-as-control-flow is gone.
- **Extracted `checkIsPhpCoreClass`**: cache logic and detection logic are now in separate methods.

## Results

Measured on arkitect's own codebase (104 files):

| | `ReflectionClass` calls |
|---|---|
| Before | 849 |
| After | 148 |
| Reduction | **-83%** |

### DIFF
--- a/src/Analyzer/ClassDescriptionBuilder.php
+++ b/src/Analyzer/ClassDescriptionBuilder.php
@@ -187,18 +187,29 @@ class ClassDescriptionBuilder
         );
     }
 
+    /** @var array<string, bool> */
+    private static array $cache = [];
+
     private function isPhpCoreClass(ClassDependency $dependency): bool
     {
-        $fqcn = $dependency->getFQCN();
+        $className = $dependency->getFQCN()->toString();
 
-        try {
-            /** @var class-string $className */
-            $className = $fqcn->toString();
-            $reflection = new \ReflectionClass($className);
+        if (!isset(self::$cache[$className])) {
+            self::$cache[$className] = $this->checkIsPhpCoreClass($className);
+        }
 
-            return $reflection->isInternal();
-        } catch (\ReflectionException $e) {
+        return self::$cache[$className];
+    }
+
+    private function checkIsPhpCoreClass(string $className): bool
+    {
+        if (!class_exists($className) && !interface_exists($className)) {
             return false;
         }
+
+        /** @var class-string $className */
+        $reflection = new \ReflectionClass($className);
+
+        return $reflection->isInternal();
     }
 }
--- a/src/Analyzer/ClassDescriptionBuilder.php
+++ b/src/Analyzer/ClassDescriptionBuilder.php
@@ -41,6 +41,9 @@ class ClassDescriptionBuilder
 
     private ?string $filePath = null;
 
+    /** @var array<string, bool> */
+    private static array $cache = [];
+
     public function clear(): void
     {
         $this->FQCN = null;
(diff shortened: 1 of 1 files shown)


src/Analyzer/ClassDescriptionBuilder.php
```diff
@@ -189,16 +192,24 @@ public function build(): ClassDescription
...
-            $className = $fqcn->toString();
-            $reflection = new \ReflectionClass($className);
+        return self::$cache[$className];
+    }
 
-            return $reflection->isInternal();
-        } catch (\ReflectionException $e) {
+    private function checkIsPhpCoreClass(string $className): bool
+    {
+        if (!class_exists($className) && !interface_exists($className)) {
```
micheleorselli: Are there any traits or enums in the core of php? If the answer is yes, then maybe we need to add also [trait_exists](https://www.php.net/manual/en/function.trait-exists.php) and [enum_exists](https://www.php.net/manual/en/function.enum-exists.php)

micheleorselli (commented):
I left one comment on the check condition. Apart from that this seems to me a quick win to improve the current situation so 👍

src/Analyzer/ClassDescriptionBuilder.php
fain182: Good catch! I checked and PHP doesn't seem to have any built-in traits, so trait_exists wouldn't change behavior today. For enums, class_exists already returns true for built-in enums (like RoundingMode), so they're already handled correctly.

That said, adding trait_exists and enum_exists is a good idea for performance: the guard is an early exit — if a symbol isn't loaded in the current process, it can't be a core class. Adding these checks extends that optimization to traits and enums, avoiding a ReflectionClass instantiation for unloaded symbols of those kinds.

PR #610, opened 2026-04-25
fix(phar): keep Symfony polyfills out of the scoper prefix

fain182:
## Summary

- The PHAR build switched to php-scoper in `f0e7bba` and started shipping a broken PHAR: running `check` fails with `Class "Normalizer" not found` (or `Call to undefined function normalizer_is_normalized()` once the class lookup is fixed).
- Root cause: the php-scoper bundled in `bin/box.phar` (Box 3.16, Feb 2022) wraps the Symfony polyfill stubs (`Resources/stubs/*.php`) and bootstraps (`bootstrap*.php`) in `namespace _PhpArkitect;` even with `expose-global-classes` / `expose-global-functions` enabled, so `Symfony\Component\String\AbstractUnicodeString` can no longer resolve `\Normalizer` / `\normalizer_normalize()` inside the PHAR.
- Add `Symfony\Polyfill` to `exclude-namespaces` for the implementation files, and list the stubs + bootstraps via `exclude-files` (they declare no namespace, so `exclude-namespaces` does not catch them).
- Add an end-to-end smoke test that runs `phparkitect.phar check` against a fixture (default progress bar + `--autoload`) so any future regression in PHAR scoping fails CI instead of slipping through the help-only smoke test.

## Test plan

- [x] Reproduce locally with the CI build environment (`composer config platform.php 8.0 && composer update --no-dev`) — confirmed the failing test commit reproduces the regression and the released `0.8.0` PHAR passes it.
- [x] Rebuild the PHAR with the fix and run `check` end-to-end on the autoload fixture and on `configMvc.php` (violation rendering also exercises the polyfill path) — both work.
- [x] Full unit test suite still green (405/405 on PHP 8.5).
- [ ] CI green on the matrix (PHP 8.0 → 8.5) for both `phar` and `smoke-test-phar` jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,6 +103,8 @@ jobs:
       matrix:
         php-versions: [ '8.0', '8.1', '8.2', '8.3', '8.4', '8.5' ]
     steps:
+      - uses: actions/checkout@v6
+
       - name: Install PHP
         uses: shivammathur/setup-php@v2
         with:
@@ -118,6 +120,12 @@ jobs:
 
       - name: "Smoke test phar"
         run: php ./phparkitect-${{ github.sha }}.phar
+
+      - name: "End-to-end test phar: run check on fixture"
+        run: |
+          php ./phparkitect-${{ github.sha }}.phar check \
+            --config tests/E2E/_fixtures/autoload/phparkitect.php \
+            --autoload tests/E2E/_fixtures/autoload/autoload.php
 
   publish_phar:
     needs: [build, smoke-test-phar]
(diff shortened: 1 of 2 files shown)


fain182:
Superseded by #611, which upgrades box to 4.7.0 and fixes the underlying php-scoper polyfill bug at the source. The `exclude-files` / `exclude-namespaces` workaround in this PR is no longer needed once #611 lands.

Closing in favor of #611. The end-to-end PHAR smoke test introduced here will be re-added on top of the box upgrade in a follow-up if not bundled into #611.

PR #615, opened 2026-05-22
added Not have attribute expression

AlessandroMinoccheri:
Summary
Adds NotHaveAttribute, the missing negative counterpart to HaveAttribute, following the same pattern used by every other expression pair in the codebase (HaveTrait/NotHaveTrait, Implement/NotImplement, etc.).

This allows rules like:
```
php$rules[] = Rule::allClasses()
    ->that(new ResideInOneOfTheseNamespaces('App\Controller'))
    ->should(new NotHaveAttribute('Deprecated'))
    ->because('deprecated controllers should be removed, not kept in production');
```

Notes
No breaking changes. No new dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -518,13 +518,18 @@ $rules[] = Rule::allClasses()
     ->because('we don\'t want to enforce immutability');
 ```
 
-### Has an attribute (requires PHP >= 8.0)
+### Has an attribute / Does not have an attribute
 
 ```php
 $rules[] = Rule::allClasses()
     ->that(new ResideInOneOfTheseNamespaces('App\Controller'))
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -518,13 +518,18 @@ $rules[] = Rule::allClasses()
     ->because('we don\'t want to enforce immutability');
 ```
 
-### Has an attribute (requires PHP >= 8.0)
+### Has an attribute / Does not have an attribute (requires PHP >= 8.0)
```
fain182: we can correct this, phparkitect needs php 8.0 to run now

fain182 (approved):
LGTM 🙂

PR #617, opened 2026-05-27
Prevent autoloader invocation when checking for PHP core classes

fain182:
## Summary

Fixes a bug where `checkIsPhpCoreClass()` could trigger the Composer autoloader during file parsing, causing a spurious "parsing error" and exit code 1 on valid PHP files.

## How we found it

Reported via [Sylius/Sylius#19003](https://github.com/Sylius/Sylius/pull/19003), where upgrading phparkitect from `^0.8` to `^1.0` broke the architectural check with:

```
❌ found parsing errors in these files:
Class "Doctrine\ODM\MongoDB\DocumentRepository" not found in file: TaxonomyBundle/Resources/config/services/integrations/doctrine/mongodb-odm.php
```

The file is a perfectly valid Symfony DI config. The error was triggered because phparkitect's own process called `class_exists('Doctrine\ODM\MongoDB\DocumentRepository', true)` while analysing it, which invoked the autoloader. Since the `mongodb` PHP extension was not installed in that CI environment, loading the class file threw a PHP `Error` that propagated out of `class_exists()` and was caught by `FileParser`'s `\Throwable` handler — turning a valid file into a parsing error.

## Fix

Pass `$autoload = false` to all `*_exists()` calls in `checkIsPhpCoreClass()`. PHP built-in classes are always pre-loaded; if a symbol is not already present in the process it cannot be a PHP internal class, so triggering the autoloader is both unnecessary and harmful.

https://claude.ai/code/session_01A9BJtoBs8od2HvUya66nq2

### DIFF
--- a/src/Analyzer/ClassDescriptionBuilder.php
+++ b/src/Analyzer/ClassDescriptionBuilder.php
@@ -203,12 +203,7 @@ class ClassDescriptionBuilder
 
     private function checkIsPhpCoreClass(string $className): bool
     {
-        // Use $autoload=false to avoid triggering the Composer autoloader for arbitrary class
-        // names encountered during file parsing. PHP built-in classes are always pre-loaded and
-        // do not need autoloading. Triggering the autoloader here can cause side-effects: if a
-        // class's file fails to load (e.g. because a required PHP extension is missing), PHP
-        // throws an Error that propagates out of class_exists() and gets mistakenly treated as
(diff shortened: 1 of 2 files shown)


micheleorselli (approved):
Imo the best way to deal with that is to not rely on reflection at all. Said that, this PR should fix the issue 🚀

PR #621, opened 2026-06-05
ignored line numbers in the baseline

### LINKED ISSUE #620: Ignore line numbers
When working with a big team on a legacy code base with millions of lines of code, having line numbers in the baseline doesn't really work well. I added --ignore-line-numbers to the generation of the baseline and when running tests but the line numbers still exist in the baseline and you still get constant merge conflicts.

Would you consider changing it to a count of error types, the same way phpstan does?

Thanks

AlessandroMinoccheri:
Fix --ignore-baseline-linenumbers not applied when generating baseline
Fixes #620 

When using --generate-baseline --ignore-baseline-linenumbers, line numbers were still written to the baseline file. 

The flag only affected runtime comparison, not serialization, causing constant merge conflicts in large teams on legacy codebases.
Line numbers are now stripped before saving when --ignore-baseline-linenumbers is passed. No existing behaviour is affected — the fix only applies when both flags are used together.

### DIFF
--- a/src/CLI/Baseline.php
+++ b/src/CLI/Baseline.php
@@ -65,10 +65,14 @@ class Baseline
         );
     }
 
-    public static function save(?string $filename, string $defaultFilePath, Violations $violations): string
+    public static function save(?string $filename, string $defaultFilePath, Violations $violations, bool $ignoreLineNumbers = false): string
     {
         if (null === $filename) {
             $filename = $defaultFilePath;
+        }
+
+        if ($ignoreLineNumbers) {
+            $violations = $violations->withoutLineNumbers();
         }
 
         file_put_contents($filename, json_encode($violations, \JSON_PRETTY_PRINT));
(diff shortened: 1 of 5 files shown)


micheleorselli (approved):
If I got the code right, with this change if a user generates a baseline with `--ignore-baseline-linenumbers` then every comparison with that baseline would (obviously) ignore line numbers, meaning that if the tool is run with `--ignore-baseline-linenumbers` or not, it would produce the same result

⚠️ Maybe we should document this behaviour and/or make explicit if the comparison is considering line numbers or not

💭 In my opinion the long term solution should be rethinking the way the baseline works. The user should not be forced to know about line numbers, it should just work. Psalm and PHPStan do use baseline, so we could see how they do it

AlessandroMinoccheri:
I agree @micheleorselli , I would suggest to merge this PR and create a new issue as an IDEA with your vision.
Is it ok for you?
@fain182  what do you think?

fain182:
I think we should add a section in the readme in which document all the configurations...
ignore baseline numbers it seems more like a configuration than an option, because is something persistent.

PR #625, opened 2026-06-17
fix: make init suggest a command instead of a wrong path

fain182:
## What

The `init` command prints `php bin/phparkitect check` after creating the config file, but that path is wrong for most users:

- **Composer install**: the binary lands in `vendor/bin/phparkitect` (Composer's default `bin-dir`), not `bin/`.
- **Phar**: it's run as `./phparkitect.phar check`.
- `bin/phparkitect` is only correct when developing inside this repo (where `config.bin-dir` is set to `bin`).

Since no single path is correct for every install method, this changes the message to the neutral `phparkitect check`, which is right regardless of how the tool was installed. The same string in the "file already exists" branch is updated for consistency, along with the E2E assertions.

### DIFF
--- a/src/CLI/Command/Init.php
+++ b/src/CLI/Command/Init.php
@@ -54,7 +54,7 @@ EOT;
 
             if (file_exists($destFilePath)) {
                 $output->writeln('<info>File</info> phparkitect.php <info>found in current directory, nothing to do</info>');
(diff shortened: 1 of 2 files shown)


AlessandroMinoccheri (approved):
LGTM

PR #631, opened 2026-06-23
Cover violation file with tests

AlessandroMinoccheri:
Adding tests for Violation.php file

### DIFF
--- a/tests/Unit/Rules/ViolationTest.php
+++ b/tests/Unit/Rules/ViolationTest.php
@@ -1,0 +1,95 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Arkitect\Tests\Unit\Rules;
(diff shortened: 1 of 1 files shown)


tests/Unit/Rules/ViolationTest.php
```diff
@@ -0,0 +1,95 @@
...
+        self::assertEquals($original->getFilePath(), $restored->getFilePath());
+    }
+
+    public function test_from_json_round_trips_without_file_path(): void
+    {
+        $data = [
+            'fqcn' => 'App\Baz',
+            'error' => 'should extend BaseClass',
+            'line' => null,
+        ];
```
fain182, line 68: here you wanted to do as above with tojson and then fromjson?

PR #659, opened 2026-07-19
build: manage psalm and box via composer in tools/ instead of committed phars

fain182:
This replaces the psalm and box phars committed in `bin/` with two small composer projects, `tools/psalm` and `tools/box`, each with its own committed lock file.

**Why.** The phars are outside every update mechanism we have: nothing watches them, so `psalm.phar` has been stuck at the version committed in #450 and today it cannot even start on PHP >= 8.4 dev machines (its embedded requirements checker stops at 8.3). As composer projects the tools are now covered by Dependabot (two new entries in `dependabot.yml`), which also means security advisories for them are surfaced like for any other dependency. As a bonus, the repo stops accumulating multi-MB binary blobs in its git history on every tool update, and tool provenance moves from "binary committed by hand" to packages resolved from Packagist with a lock file.

**Why this still avoids the problem #450 solved.** Psalm was moved out of `require-dev` because its own dependencies (e.g. its `nikic/php-parser`) conflict with arkitect's. The `tools/` directories keep that property: each tool has an isolated vendor tree, nothing is shared with the root project.

**What changes.** Tool versions are unchanged (psalm stays on 5.26, box was already at the latest 4.7), `make psalm` / `make phar` / `make build` keep working the same way (they run `composer install -d tools/...` first, a fast no-op when up to date), and the CI psalm step still runs on the 8.0 job. Platform pins in the tools' `composer.json` keep the lock files resolved for the oldest runtime each tool runs on in CI (8.0 for psalm, 8.2 for box). The `.gitignore` `composer.lock` pattern is anchored to the root so the tools' locks are tracked while the root lock stays uncommitted by design.

The upgrade to Psalm 6 is deliberately left out of this PR and will follow as a separate one (branch `chore/upgrade-psalm-6`), since it brings its own config decisions and code fixes.


.github/workflows/build.yml
```diff
@@ -37,7 +37,9 @@ jobs:
       id: composer-cache
       uses: actions/cache@v6
       with:
-        path: vendor
+        path: |
+          vendor
+          tools/psalm/vendor
```
AlessandroMinoccheri, line 42: Why only tools/psalm/vendor but not tools/box/vendor? box's composer install will re-run uncached on every phar-build job. Low impact imo

fain182: I agree, box should be here too 🤔

fain182: Oh, i got it, this cache step is in the build job, which never installs box (only psalm). And the phar job where box lives has no cache restore, and jobs don't share a filesystem — so nothing would be cached either way.

Issue #636, opened 2026-06-30
--ignore-baseline-linenumbers can report the wrong violation as new


fain182:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Library Version | 1.1.1
| PHP version     | 8.3

#### Summary

When using `--ignore-baseline-linenumbers`, adding a *new* violation to a class that already has other baselined violations of the same rule can cause `Violations::remove()` to flag the wrong line as "new" — while silently absorbing the actually-new violation as if it were already known. The report doesn't reflect what was really added/changed in the code.

#### Current behavior

`Violations::remove()` (`src/Rules/Violations.php`) matches current violations against baseline violations by `(fqcn, normalized error message)` only, ignoring the line number as the option name implies. Both lists are pre-sorted by `fqcn` then by `line` (`Violations::sort()`, called in `Runner::doRun()`), and the matching loop pairs them up **positionally**, in sorted order — not by stable identity.

This means: if a class already has N baselined violations of the same rule, and a new violation of the same rule is introduced at a line that sorts *before* one of the existing ones, the position-based pairing shifts and an existing (untouched) violation can end up unmatched and reported as "new", while the actual new one is paired away with an older baseline entry and never reported.

Concrete trace — class `Foo` already has 3 baselined violations of the same rule at lines 10, 20, 30. A 4th violation of the same rule is introduced at line 15 (between the first two):

- current run, sorted by line: `[10, 15(new), 20, 30]`
- baseline, sorted by line: `[10, 20, 30]`

Pairing proceeds index by index:
- `10` ↔ `10` → matched (correct)
- `15` (new) ↔ `20` (pre-existing, untouched) → matched — the genuinely new violation is silently discarded as "already known"
- `20` (pre-existing) ↔ `30` (pre-existing) → matched
- `30` (pre-existing) ↔ *nothing left* → reported as the "new" violation

The tool reports line 30 — code that was never touched — as the new violation, while the line actually introduced (15) produces no output at all.

This is a different failure mode from #428/#430: that fix addressed baseline entries not being fully cleared when the violation *set* was unchanged. This one is about misattribution when the set *does* change, specifically when a new violation's line sorts earlier than existing siblings in the same fqcn.

#### How to reproduce

1. Create a class with 3 dependency violations of the same rule (same fqcn, same rule message) at increasing line numbers, baseline them with `--generate-baseline`.
[...]

PR #666, opened 2026-07-26
feat: match baseline violations without asking the user how

### LINKED ISSUE #662: Baseline matching should not require the user to pick a strategy (line numbers vs. not)
## Context

Today the user has to choose how baseline violations are matched against the current run, via the `--ignore-baseline-linenumbers` / `-i` flag (and the corresponding `ignoreBaselineLinenumbers()` in the config file). This drives two separate matching code paths in `Violations::remove()` and `Violations::countUnmatchedIn()`:

- **with line numbers** (default): a violation matches a baseline entry only if FQCN, line number, file path and the (stable part of the) error message match. Any change *above* the offending line shifts the line number and the check starts failing on an already-known violation.
- **without line numbers** (`-i`): line numbers are dropped from the match, so unrelated edits don't reopen a baselined violation — but two violations of the same kind in the same class become indistinguishable.

## Problem

The matching mechanism is an **implementation detail**, and it doesn't make sense that the user has to understand and choose it. It's not a decision about *what* to enforce (which is the user's job), it's a decision about *how* the baseline is compared (which is ours). The current situation:

- forces users to learn the trade-off between the two modes;
- makes baselines behave differently across projects/CI depending on a flag;
- the default (line-number-sensitive) produces false failures on unrelated edits, which is a common source of confusion.

## Proposal

There should be a **single matching strategy that works well for everyone**, with no user-facing knob. The line-number sensitivity in particular should not be something the user opts into.

Approaches to evaluate:

- **Hash of the line content** instead of the line number. Anchoring a baseline entry to a hash of the offending line's source (optionally normalized for whitespace) makes matching stable against edits elsewhere in the file, while still distinguishing multiple violations of the same kind in the same class — which is exactly what the two current modes each fail at.
- Fuzzy line matching (nearest line with the same content) as a lighter-weight alternative.
- Keep FQCN + stable error key as the primary identity and use the line only as a tie-breaker.

## Backwards compatibility

- `--ignore-baseline-linenumbers` / `ignoreBaselineLinenumbers()` would be **deprecated** (kept as a no-op with a notice, then removed in a future major).
- Existing baseline files may need a one-time regeneration; `generate-baseline` should produce the new format.

## Related

- #636 — bug report on `--ignore-baseline-linenumbers` reporting the wrong violation as new. It explores the matching-identity solution space in depth (hash of source line, enclosing member, count-based grouping à la PHPStan/ESLint, git-diff à la golangci-lint). This issue is the DX counterpart: rather than *fixing* the flag's behaviour, it argues the flag (the user-facing choice) should not exist at all, and a single robust strategy should be the default for everyone. Whatever identity scheme #636 lands on is a strong candidate for that single strategy.
- #649 — `prune-baseline` command; touches the same area since pruning would refresh stale line numbers.

## Notes

Relevant code: `src/Rules/Violations.php` (`remove()`, `countUnmatchedIn()`, `compareViolations()`, `extractViolationKey()`), `src/CLI/Baseline.php`, and the option wiring in `src/CLI/Command/CommonOptions.php`.

fain182:
Choosing between matching the baseline with or without line numbers is a decision about *how* the baseline is compared, not about *what* to enforce — and both answers were wrong somewhere. The default reopened a known violation after any edit above it; `--ignore-baseline-linenumbers` could not tell two violations of the same rule in the same class apart, and could report the wrong one as new (#636).

Matching is now a single strategy with no knob, in two passes:

1. violations still sitting where the baseline recorded them pair first;
2. among what is left, class and reported problem alone decide.

So an edit above a violation moves it without making it new, identical violations in one class stay distinct, and when a violation is added among moved ones it is the added one that gets reported. The `#636` scenario (three baselined violations at 10/20/30, a fourth added at 15) is covered by a test.

When a group of identical violations both moved and grew in the same run, which of them is reported as new is a guess — the count is not. That ambiguity is inherent, and it is documented in the README rather than papered over.

**Deprecation**: `--ignore-baseline-linenumbers` / `ignoreBaselineLinenumbers()` have no effect and print a notice, whether they come from the CLI or from `phparkitect.php`. They will be removed in the next major. `generate-baseline` always writes line numbers now; baselines stored without them keep working and need no regeneration, and `prune-baseline` still preserves that format.

Pairing stays linear: a synthetic 20k-entry baseline where every violation moved matches in 0.04s, and 2000 violations of the same rule in a single class in under 0.01s.

Built on #665, which put all matching in `Violations::matchAgainst()` — this PR is mostly the removal of the boolean from it.

Closes #662
Refs #636

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ Every setting can be passed as a CLI option or set via the corresponding `Config
 | `--autoload` | `-a` | `autoloadFilePath()` | Autoload file to load before running. Required for all Phar runs. |
 | `--use-baseline` | `-b` | `baselineFilePath()` | Baseline file path for ignoring known violations. |
 | `--skip-baseline` | `-k` | `skipBaseline()` | Skips the default baseline even if present. |
-| `--ignore-baseline-linenumbers` | `-i` | `ignoreBaselineLinenumbers()` | Matches baseline violations without checking line numbers. |
+| `--ignore-baseline-linenumbers` | `-i` | `ignoreBaselineLinenumbers()` | **Deprecated**: has no effect, baseline matching already tolerates moved violations. |
 | `--config` | `-c` | — | Configuration file to load (default: `phparkitect.php`). |
 | `--verbose` | `-v` | — | Prints every parsed file instead of the progress bar. |
 | — | — | `skipParsingCustomAnnotations()` | Disables custom DocBlock annotation parsing (enabled by default). |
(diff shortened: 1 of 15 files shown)


AlessandroMinoccheri (approved):
LGTM 👍

PR #677, opened 2026-09-21
Fix namespace matching: ancestors in DependsOnlyOnTheseNamespaces, siblings in ResideInOneOfTheseNamespaces

### LINKED ISSUE #675: Ancestor is considered to be in the same namespace, breaking DependsOnlyOnTheseNamespaces rule?
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | yes
| Library Version |1.3.0
| PHP version     | 8.3.33

#### Summary

Hi, thanks for this tool that I'm using in several of my projects. Today I've experienced something strange which I've investigated (a PoC will follow).
The issue is in the `DependsOnlyOnTheseNamespaces` rule behavior. 
Looking from the outside I was expecting that only classes into whitelisted namespaces should be considered "ok" as dependencies.
With my surprise, however, I've stumble upon this particular case:

```
// Foo\Bar\FooBar.php
[...]
use Foo\Collaborator.php // Pay attention to this line

public function __toString(): string { ... } // More to follow with this one
[...]

// arkitect.php
[...]
$rules[] = Rule::namespace(['Foo\Bar'])
        ->should(new DependsOnlyOnTheseNamespaces(['FizzBuzz]))
        ->because([...]);
[...]
``` 

#### Current behavior

I was expecting `Collaborator.php` to be reported as an unwanted dependency which is not the case at the moment.
Diggining into the code I can clearly see where's the logical issue (from my standpoint).
At https://github.com/phparkitect/arkitect/blob/main/src/Expression/ForClasses/DependsOnlyOnTheseNamespaces.php#L42-L44 we're checking for a dependency to match certain namespace but matching https://github.com/phparkitect/arkitect/blob/main/src/Analyzer/PatternString.php#L21-L27 is done by a substring of the pattern. This means that every "ancestor" class in the same "sub-namespace" of the class under check is considered to be in the same namespace.

Just two consideration:

- An ancestor class should not be considered in the same namespace as, formally, the namespace isn't the same.
- I would object that the line highlighted above in `DependsOnlyOnTheseNamespaces` should be there at all: I can understand - if I'm doing properly - that if a class is _actually_ in the same namespace of the one under check should be considered a non violation but if I'm whitelisting some other classes, I would expect the same namespace to be explicitly whitelisted (this is an opinionated statement by the way) in order to accept any class in that namespace as a dependency.

If I get rid of the lines highlighted in `DependsOnlyOnTheseNamespaces`, I get errors about `Foo\Bar\string` as if the tool isn't capable of infer native types and the like.

In conclusion, I can bypass the bug adding something like 

```
->andShould(new NotDependsOnTheseNamespaces(['Foo']))
``` 

to have the evidence of unwanted dependencies but I think it could become cumbersome somehow if more granular classes must be accepted.

#### How to reproduce

You can check this PoC https://github.com/DonCallisto/arkitect/tree/namespaces_poc by comparing it with `main` on my fork.
You can also check this GHA run https://github.com/DonCallisto/arkitect/actions/runs/35356156536 which confirms my suspects; can't say if that's a bug tho.

#### Expected behavior

I would expect `Collaborator.php` to be reported as stated above

fain182:
Fixes #675, plus a second bug of the same family found while working on it.

Both come from the same category error: passing a concrete namespace where a **user-written pattern** was expected. `PatternString`'s prefix semantics are correct for a pattern — `App\Domain` should reach `App\Domain\Order\Item` — but applied to a value they silently become "is an ancestor of", and `namespaceMatches()` compares against the FQCN, not the namespace its name suggests.

### 1. `DependsOnlyOnTheseNamespaces` allowed every ancestor namespace (#675)

The rule skips dependencies in the class' own namespace so that namespace need not be whitelisted. The check was `$theClass->namespaceMatches($dependency->getFQCN()->namespace())`, which boils down to `str_starts_with($classFqcn, $pattern.'\')` — so every ancestor matched too. A class in `Foo\Bar` depending on `Foo\Collaborator` was accepted whatever the allow-list said. Child namespaces were already reported, which is the asymmetry the issue describes.

Now the two namespaces are compared exactly. The root namespace is excluded from the allowance, so user-defined classes sitting there keep being evaluated like any other, as they have been since 1.0.0.

### 2. `ResideInOneOfTheseNamespaces` selected sibling namespaces

The expression appended `'*'` to every namespace, forcing `PatternString` down its `fnmatch` branch where `*` crosses namespace separators. `ResideInOneOfTheseNamespaces('App\Domain')` therefore also selected `App\DomainEvents\Foo`. This one sits in the rule **selector**, so it silently widened which classes a rule applies to, with no signal to the user.

The trailing `*` is load-bearing — it is what lets `Food\*\Roots` reach `Food\Vegetables\Roots\Orange\Carrot` — so descending stays, but it now has to cross a separator.

### Behaviour change

Verified by running `main` and this branch side by side; everything not listed is unchanged.

| constraint, on the class it is checked against | before | after |
|---|---|---|
| `DependsOnlyOnTheseNamespaces(['FizzBuzz'])` — `Foo\Bar\FooBar` depends on `Foo\Collaborator` (ancestor) | no violation | **violation** |

| selector, on the class it is matched against | before | after |
|---|---|---|
| `ResideInOneOfTheseNamespaces('App\Domain')` — `App\DomainEvents\Foo` (sibling) | selected | **not selected** |
| `ResideInOneOfTheseNamespaces('Food\*\Roots')` — `Food\Vegetables\RootsAndTubers\Carrot` (sibling, through a wildcard) | selected | **not selected** |
| `ResideInOneOfTheseNamespaces('')` — any class | selected | **not selected** |

The first table surfaces violations that were hidden. The second changes which classes a rule applies to, and reaches `Rule::namespace()` and the whole `Architecture` DSL.

https://claude.ai/code/session_01E5jZnK8GCeQUjggeHJpqYA


fain182:
replaced by #678

